PR #149, opened 2026-04-25
OP-120: op-work agent registration with session id and conflict detection

earchibald:
## Summary

- `op-work` gains optional `agent=<id>`, `agent_session=<sid>`, and `force=true` params so agents started outside the plugin's `op:open-agent` UI can self-register on the issue note.
- Plugin writes `agent:` and `agent_session:` frontmatter, detects mismatched-agent and mismatched-session conflicts, and refuses to overwrite without `force=true`.
- Status flip + TASKS seed unchanged when new params are absent — fully backward-compatible.
- Skill teaches agents to surface conflicts to the user instead of force-overwriting; `reference/schema.md` documents the new fields and their soft-lock semantics.

## Test plan

- [x] `npx vitest run` — 427/427 pass (new `workIssue.test.ts` covers fresh register / idempotent / agent-mismatch / session-mismatch / force / session upgrade / agent-only / empty-args; `cliHandlers.test.ts` extended for `parseWorkParams`).
- [x] Live smoke test against OP-120 itself: fresh register → idempotent → session conflict (refused) → force override (succeeded). Frontmatter on disk reflects each transition.
- [ ] Reviewer to double-check `clearAgentOnIssue` (`openAgent.ts`) and `runResolve` (`resolve.ts`) now also clear `agent_session` for symmetry.


earchibald:
@copilot please perform an adversarial code review of this PR. Specifically pressure-test:

1. **Conflict logic correctness** in `workIssue.ts` — the four cases (no existing / matching / agent-mismatch / session-mismatch) and the `force=true` override. Are there edge cases the unit tests miss? E.g. what happens when only `agent` is passed but the existing issue has both `agent` and `agent_session`?
2. **Backward compatibility** — does the new behavior change anything for callers that don't pass `agent`/`agent_session`? The plugin's own `op:open-agent` UI still calls `workIssue` without args; verify that path is unchanged.
3. **Concurrency / TOCTOU** — `workIssue` reads then writes `fm.agent`/`fm.agent_session` inside `processFrontMatter`. Could two parallel `op-work` calls both see no existing registration and both claim?
4. **Schema/skill drift** — `SKILL.md` and `reference/schema.md` describe the new soft-lock semantics. Do they match the implementation? Any lies or omissions?
5. **Cleanup symmetry** — `resolve.ts` and `clearAgentOnIssue` (`openAgent.ts`) now both clear `agent_session`. Are there other paths that should also clear it that I missed (e.g. crash-recovery hooks, session-end handlers)?
6. **Param validation** — `parseWorkParams` accepts arbitrary non-whitespace strings for `agent`. Should there be an allowlist? What about `agent_session` length / character restrictions?

Be skeptical and concrete. Cite file:line. Rank findings by severity.

Issue #150, opened 2026-04-25
Make op-obsidian feel great: UX/hotkey/transition improvements brainstorm


earchibald:
Tracked as op issue **OP-149**.

## Scope

- [ ] Buttons and inline actions in issue notes (status bumps
- [ ] launch agent
- [ ] append commit
- [ ] resolve)
- [ ] Sidebar information density: density modes
- [ ] group-by-project
- [ ] per-flow filters
- [ ] multi-select bulk actions
- [ ] Hotkey-driven keyboard-only flow: pick issue
- [ ] launch
- [ ] switch
- [ ] resolve without touching mouse
- [ ] Smoother Obsidian↔terminal transitions: focus-bounce on iTerm close
- [ ] hide-vs-quit
- [ ] optional SwiftUI shim
- [ ] Agent badge persistence: keep linkage visible while session is open even after resolve
- [ ] Command-palette ergonomics: shorter names
- [ ] grouping
- [ ] fuzzy aliases
- [ ] smart-default verbs
- [ ] Status-bar + ribbon enhancements: live agent count
- [ ] current-issue chip
- [ ] quick-launch dropdown
- [ ] Note-level decorations: agent activity indicator at top of issue
- [ ] last-commit chip
- [ ] PR status badge
- [ ] Quick-capture: 'op: new from selection / from clipboard' to hand-off context to a fresh issue in one keystroke
- [ ] Errors and notices: actionable Notices with one-click 'Open settings' / 'Open log' buttons
- [ ] Onboarding and discoverability: first-run guided tour or in-vault demo project
- [ ] Adversarial copilot review of the resulting spec

PR #151, opened 2026-04-25
OP-149: feel-great UX brainstorm + design spec

earchibald:
## Summary

Design spec only — no code changes. Proposes 13 themed UX improvements for op-obsidian, ordered against one bar: *a power user should be able to drive an issue from "thought" to "merged" without taking their hands off the keyboard, and never feel a jerk between Obsidian and the agent terminal.*

Resolves: spec deliverable for OP-149 (https://github.com/earchibald/obsidian-projects/issues/150).

Spec lives at: `docs/specs/OP-149-feel-great-ux-design.md`

## What's in the spec

13 themed sections, each with current state, proposal, why it feels better, cost/risk, verdict (RECOMMEND / HOLD / REJECT). Headline ideas:

1. **Resumability** — `op: resume last` + sidebar header chip + 5s tmux liveness probe. Adversarial pass surfaced this as the missing lever; now leads.
2. **One contextual primary chip** above the H1 of every issue note (CM6 widget + Reading-mode mirror). Not a wall of buttons.
3. **Opinionated hotkey preset** + `op: pick & act` modal (Raycast/Omnisearch single-modal pattern).
4. **Terminal transitions** — `open -ga iTerm`, post-create activate-bounce, document iTerm "Hide on close tab". SwiftUI shim explicitly rejected.
5. **Agent badge persistence across resolve** — keep `agent:` if tmux window is alive; new `op: detach agent`.
6. **Sidebar polish** — keyboard nav, density pref, right-click menus. Multi-select + per-flow filters rejected.
7. **Palette ergonomics** — hide `op-dev:*` behind a setting, trim wordy command names.
8. **Quick-capture** — `op: new from selection / clipboard`.
9. **Actionable Notices** — clickable `[Open]` / `[Retry]` actions in the high-value Notices.
10. **Onboarding** — first-run `Projects/_op-readme.md` with working Apply-preset chip and demo-project scaffolder.
11. **Inline GH/PR status** — strip below the chip showing last commit + PR + GH issue state.
12. **Right-click row menus** in the sidebar (drag-to-resolve rejected as too easy to misfire).
13. **Settings tab UX cleanup** — split 11 H2 sections into Daily/Advanced + in-tab search + per-section collapsibles + targeted re-renders.

Plus an explicit no-pile (§14) covering things that look good but aren't (status-bar widget, dashboard view, animated transitions, mobile parity, etc.) and an architecture diagram (§15) showing 9 new single-purpose modules layered on existing ones.

Each RECOMMEND becomes a follow-up issue. Sequenced 10 PRs in §16, totaling ~1,600 LoC touched, no breaking changes.

## Diagrams included

Mermaid: at-a-glance today-vs-proposed flow, sidebar liveness probe, chip-state decision tree, focus-stack sequence (showing where each terminal-transition fix intervenes), resolve-path liveness gate, new-vs-existing module map. ASCII: sidebar header chip, note-level chip + overflow menu, pick-and-act modal, proposed settings tab shape.

## Test plan
- [ ] Mermaid diagrams render in GitHub PR view (verify after push)
- [ ] No ambiguity — each section has a verdict, no "TBD"
- [ ] Each RECOMMEND is sized for one PR (cross-checked in §16)
- [ ] Adversarial Copilot review surfaces concrete pressure points (see comment thread)

### DIFF
--- a/docs/specs/OP-149-feel-great-ux-design.md
+++ b/docs/specs/OP-149-feel-great-ux-design.md
@@ -99,7 +99,13 @@ flowchart TD
 
 **Why it feels better.** The single highest-frequency op-obsidian action — figuring out what to work on next — becomes one keystroke. The sidebar stops lying about which agents are alive.
 
-**Cost / risk.** Small. Recency log is ~30 LoC. Liveness probe is one `tmux` exec on a 5s interval gated on visibility — negligible. The probe is the only new platform call; on non-macOS it no-ops (tmux available everywhere we ship).
+**Cross-vault note.** `data.json` lives at `.obsidian/plugins/op-obsidian/data.json` — one file per vault. A multi-vault user's recency log in Vault A never appears in Vault B. Per-vault recency is not only correct; it's the only thing possible with this storage layer.
+
+**`data.json` corruption recovery.** The existing `mergeSettings()` call passes the parsed result through `Object.assign({}, DEFAULT_SETTINGS, stored)` — if `recent` is missing or non-array it gets replaced by the default (`[]`). This handles missing-key gracefully. It does **not** handle the case where `data.json` itself is unparseable JSON (full file corruption). For the `recent` array specifically, add a guard at the read site: `settings.recent = Array.isArray(settings.recent) ? settings.recent : [];` — this is two extra lines and ensures that any corruption of just this field is self-healing without touching the rest of the settings recovery path.
+
+**Liveness probe exec timeout.** `tmux list-windows` under normal conditions takes 1–5ms. Under pathological conditions (tmux zombie process, NFS home directory, kernel scheduling hiccup), the exec can hang indefinitely. The probe runs on the main thread via `child_process.execSync` and will block the sidebar render if it hangs. Fix: use `execSync` with an explicit `timeout` option: `execSync('tmux list-windows -t op-agents -F \'#W\'', { timeout: 500 })` — 500ms is generous for any healthy tmux and will throw if exceeded, which the caller already handles as "tmux unavailable". This prevents the render thread from stalling.
+
+**Cost / risk.** Small. Recency log is ~30 LoC. Liveness probe is one `tmux` exec on a 5s interval gated on visibility — negligible. The probe is the only new platform call; on non-macOS it no-ops (tmux available everywhere we ship). The corruption-guard and exec-timeout together add ~5 LoC.
 
 **Verdict.** `RECOMMEND` — leads. Highest leverage per LoC in the spec.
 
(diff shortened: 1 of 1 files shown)


earchibald:
@copilot — adversarial design review please. This PR adds a 600-line design spec with no code; I want pressure on the *thinking*, not on prose. Please push hard on the points below and propose concrete edits where you disagree.

**Concrete pressure-test prompts:**

1. **§1 Resumability — recency log scope.** I propose appending `{issueId, at}` to `data.json → recent: [...]` (cap 25). Failure modes: (a) `data.json` corruption — does the existing `mergeSettings()` recover, or do we need a separate file? (b) Multi-vault users: does cross-vault recency make sense, or is per-vault correct? (c) The 5s liveness probe runs `tmux list-windows`; under what conditions does this exec hang or take >100ms, and does that block the sidebar render thread?

2. **§2 Note-level chip — CM6 widget vs. post-processor parity.** The spec proposes both a CM6 widget (Live Preview) and a post-processor mirror (Reading mode). When the user toggles between modes mid-edit, does the widget tear down cleanly without leaving stale event listeners? Also: how does the chip behave in the *unstyled "source mode"* (raw editor)? My current answer is "doesn't render" — is that the right call, or will users complain?

3. **§3 Hotkey preset — collision strategy.** I waved at "best effort vs all or nothing" in §17. What's the actual right behavior when `⌘⇧L` is already bound to the user's note-export plugin? Failing closed (skip) leaves a partial preset; failing open (clobber) destroys user state. The spec doesn't decide — should it?

4. **§5 Badge persistence — the riskiest section.** I add a tmux liveness check before `clearAgentOnIssue` in the resolve path. Pressure points:
   - What if `tmux has-session` returns success but `list-windows` fails because the window was killed *between* the two calls? (TOCTOU.)
   - What if the user runs `op-resolve` for OP-72 from the palette while OP-72's agent is mid-tool-call and the SessionEnd hook is about to fire? Race.
   - The proposed Notice "Agent session still attached — `agent:` field kept" — is this *actionable*? It tells the user a thing happened but not what to do next. Should it have an `[Open agent]` link?

5. **§4 Terminal transitions — `open -ga iTerm` claim.** I claim this launches iTerm without activating. Verify: does this actually hold when iTerm is already running (no-op) vs. cold-start (where `-g` matters)? And the post-create AppleScript `tell application "Obsidian" to activate` — does that race with iTerm's tab-create animation, leaving Obsidian focused but iTerm visually mid-fade?

6. **§13 Settings restructure — smoke test risk.** CLAUDE.md has a settings smoke-test recipe that asserts on `.op-project-order__item` count via `app.setting.openTabById`. If the section is now inside a `<details>`, are the items still in the DOM (yes, `<details>` doesn't unmount children) — but is `getBoundingClientRect` going to misbehave in some way that affects the existing drag-reorder code? Should the spec call out that the drag handlers need re-testing inside collapsed `<details>`?

7. **The §14 no-pile — am I wrong about anything?** Status-bar widgets, dashboard view, multi-select, drag-to-resolve, mobile parity — pick the one you most disagree with and argue for it.

8. **Section ordering claim in §16.** I sequence Resumability (§1) at PR #4, after Hotkey preset (§3). But the §1 chip would benefit from being clickable via the preset's bound key. Should the order be inverted?

9. **Sizing claims.** §16 puts ranges like "~30 LoC", "~200 LoC", "~400 LoC". Pick one you think is materially off and say why.

10. **Anything I missed.** What's the obvious-in-hindsight UX win I left out? The adversarial subagent flagged resumability and a few others before I wrote this — your turn to be that voice.

Specific edits welcome — push commits to this branch and I'll review.

earchibald:
Three inline comments from your review weren't fully covered by fc7ae6b's bulk integration — landed in a69fc29:

1. **§1 — `tmux | grep` shell pipeline** → replaced the example with `execFile(['list-windows', '-t', 'op-agents', '-F', '#W'])` + exact-match parse via the existing `tmuxWindowName()` helper. No shell, no substring trap (the codebase already documents the OP-3-vs-OP-31 `grep -Fxq` rationale in `terminalLaunch.ts`).

2. **§1 — non-macOS platform language** → spec now states macOS-only explicitly (matching `launchInTerminal`'s actual behavior), and describes how the recency log degrades on other hosts: still works for `op: resume last` (opens last issue's note), just no agent re-attach and no liveness demotion of badges.

3. **§3 — `app.hotkeyManager.customKeys` is internal** → mandated a `try`/fallback path: degraded mode shows a copy-to-clipboard JSON snippet + an `[Open Hotkeys settings]` button + a Notice listing the seven commands to bind. Feature never hard-fails on Obsidian API drift.

Spec is now in its final state for this issue. Total of 13 themed sections (12 RECOMMENDs + 1 with mixed verdicts), 3 mermaid diagrams, 3 ASCII mockups, explicit no-pile, 10-PR sequence at ~1,810 LoC. Each RECOMMEND becomes its own follow-up issue.

Thanks for the adversarial pass — the TOCTOU/race-resolution call on §5, the `<details>`-drag bug, the status-bar reframe, and the cross-issue keyboard navigation idea materially improved the spec.

PR #168, opened 2026-04-25
OP-158: command palette — hide op-dev:*, trim names

earchibald:
## Summary

§7 of OP-149 (`docs/specs/OP-149-feel-great-ux-design.md`). First PR in the spec rollout (per OP-165 §16, "Recommended order" #1).

- New default-off setting `Show developer commands in palette` (under a new **Advanced** section in the op settings tab). Gates the four `op-dev:*` registrations (`op-debug-agent-launch`, `op-dump-store`, `op-rebuild-store`, `op-install-agent-hooks`) behind a single `if (this.settings.developer.showDevCommands)`. Reload required to take effect — Obsidian's `addCommand` has no `removeCommand` companion; setting description calls this out.
- Trims four wordy command `name:` fields — IDs unchanged so any existing user hotkey bindings keep working:

  | id | before | after |
  |---|---|---|
  | `op-open-agent-pick` | op: open agent (pick at runtime) | op: open agent (pick) |
  | `op-open-agent-plan` | op: open agent for issue in PLAN MODE | op: open agent (plan mode) |
  | `op-open-agent-plan-pick` | op: open agent in PLAN MODE (pick at runtime) | op: open agent (plan mode, pick) |
  | `op-migrate-links` | op: migrate legacy parent_issue/subissues to parent/children | op: migrate legacy issue links |

  The bonus fourth trim (`op-open-agent-plan-pick`) keeps the trio of plan/pick names uniform — without it the lone unaltered name was the longest of the four.
- Defers `op: resume last` (§1) and `op: pick & act` (§3) — those commands belong to OP-150/OP-152 (later in the §16 rollout) and self-register when their plumbing lands.
- Bump `0.45.0 → 0.46.0` (minor — additive, user-facing) per WORKFLOW.md.

## Test plan

- [x] `npm test` in `plugins/op-obsidian/` — 39 files / 434 tests passing, including new `developer.showDevCommands` default + override + reject-non-boolean cases.
- [x] `node scripts/bump-version.mjs minor` — manifest, package.json, plugin.json all at `0.46.0`; `main.js` rebuilt and fresher than `manifest.json` (the OP-105 guardrail check).
- [x] Synced build to active vault and reloaded plugin.
- [x] Probed `app.commands.commands` with toggle off: 22 op-obsidian commands, no `op-dev:*` visible; the four trimmed commands have the new short names.
- [x] Probed with toggle on: 26 commands (4 dev commands appear).
- [x] Settings UI smoke-test: `app.setting.openTabById("op-obsidian")` renders the new `Advanced` H2 and the `Show developer commands in palette` toggle binds.
- [x] `obsidian dev:console` empty after both reloads.

## Adversarial review

Required (per WORKFLOW.md — touches palette + settings public surface, fails bypass criterion 4). Will request `@copilot` review with concrete pressure-test prompts after this PR opens.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.45.0",
+  "version": "0.46.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.45.0",
+  "version": "0.46.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/main.ts
+++ b/plugins/op-obsidian/src/main.ts
@@ -605,6 +605,9 @@ export default class OpPlugin extends Plugin {
     }
 
     // Install hooks in the background so SessionEnd reports land reliably.
+    // Intentionally ungated — this is the silent background path that must
+    // always run regardless of showDevCommands. Only the *palette command*
+    // (above) is gated; removing it from the palette does not disable hooks.
     void this.runInstallAgentHooks(false);
 
     const resolveFlags = {
--- a/plugins/op-obsidian/src/main.ts
+++ b/plugins/op-obsidian/src/main.ts
@@ -389,19 +389,19 @@ export default class OpPlugin extends Plugin {
 
     this.addCommand({
       id: "op-open-agent-pick",
-      name: "op: open agent (pick at runtime)",
+      name: "op: open agent (pick)",
       callback: () => this.runOpenAgentCommand(true),
     });
 
     this.addCommand({
       id: "op-open-agent-plan",
-      name: "op: open agent for issue in PLAN MODE",
+      name: "op: open agent (plan mode)",
       callback: () => this.runOpenAgentCommand(false, "plan"),
     });
 
     this.addCommand({
       id: "op-open-agent-plan-pick",
-      name: "op: open agent in PLAN MODE (pick at runtime)",
+      name: "op: open agent (plan mode, pick)",
       callback: () => this.runOpenAgentCommand(true, "plan"),
     });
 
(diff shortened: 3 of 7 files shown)


earchibald:
@copilot please review this PR with these concrete pressure-test prompts in mind:

1. **Live-toggle correctness.** The toggle is intentionally reload-required because Obsidian's `addCommand` has no `removeCommand` companion. Verify there's no path where flipping `developer.showDevCommands` from true → false leaves a registered `op-dev:*` command callable (or vice versa pre-reload). Confirm the description text steers users correctly — does "Reload the plugin after changing" cover both directions? Should the UI add a "Reload now" button or a Notice nudge?

2. **Hotkey backwards-compat.** All four trimmed names share the same command IDs as before. Confirm no `app.hotkeyManager.customKeys` entries reference the old display names (they shouldn't — Obsidian binds by id). Are there any settings-export / -import flows where renaming `name:` could break round-tripping?

3. **`mergeSettings` defensive parsing for `developer`.** The new branch follows the same shape as `agents`/`github` (`object` check, then per-field boolean check, falls back to default). Are there inputs (`null`, `Array`, prototype-pollution `__proto__`, `Symbol`-keyed objects) where the new branch behaves differently from the existing ones?

4. **`op-install-agent-hooks` gating asymmetry.** The hook registration call (`runInstallAgentHooks(false)`) on plugin load is **not** gated — only the user-facing palette command is. Confirm that's the right call: `runInstallAgentHooks(false)` is the silent background-install path that needs to keep running so SessionEnd reports land. Gating it would silently break the hook-install pipeline. Worth a comment in the code?

5. **Bonus fourth trim (`op-open-agent-plan-pick`).** Spec listed three trims; this PR adds a fourth for consistency. Is there any user the spec-listed three trims would have left in a *worse* state than this trims-all-four version?

6. **Default false rationale.** `developer.showDevCommands` defaults to false. For *existing* users who have `op-dev:*` in their palette today, this PR silently hides those commands on next reload. Acceptable? Migration note? Or should a one-time Notice fire on first load post-upgrade if any `op-dev:*` command had a custom hotkey bound?

7. **Test surface.** `settings.test.ts` covers default + override + non-boolean. Anything missing — e.g. nested-key validation, `developer = null`, deep-clone identity (`out.developer !== DEFAULT_SETTINGS.developer`)?

8. **Build/version lockstep.** Bump went `0.45.0 → 0.46.0` across `manifest.json`, `package.json`, `.claude-plugin/plugin.json` via `scripts/bump-version.mjs minor`. Confirm the three files are all at `0.46.0` and `main.js` is fresher than `manifest.json`.

9. **No breaking changes.** Per WORKFLOW.md this is `minor`. Sanity-check that no public surface (CLI, URI handlers, slash commands, frontmatter fields) is removed or renamed; only `name:` strings change.

10. **What's missing.** Did the original spec §7 ask for anything that didn't make it into this PR (and shouldn't be deferred)? E.g. was there meant to be a "show all commands" debug toggle for plugin authors, or a way to surface `op-dev:*` ad-hoc without flipping the persistent setting?

PR #170, opened 2026-04-25
OP-160: actionable Notices — inline action links on high-value Notices

earchibald:
## Summary

- Migrates four high-value op Notices to `DocumentFragment` form with inline `<a>` action links via a new `showActionableNotice()` helper. Sticky (`duration: 0`) when actions are present so clicks aren't lost; click dismisses.
- Adds a one-shot startup probe (`surfaceStaleAgentBadgesOnStartup`) that surfaces issues with `agent:` set but no live tmux window — currently silent — with a `[Clear badge]` action. Continuous liveness probing remains OP-149 §1's scope.
- New errors from `gh issue close` are written to a vault scratch note (`Projects/_scratch/op-last-error.md`) so `[Open log]` is vault-native rather than dev-console-only.

### Migrated call sites

| Trigger | Before | After |
|---|---|---|
| Resolve succeeded | `op: OP-72 → resolved (3 tasks trashed)` | `op: OP-72 resolved (3 tasks trashed) · [Open]` |
| `op-new` succeeded | `Created OP-149` | `op: created OP-149 · [Open] · [Start agent]` |
| `gh issue close` failed | `op: gh issue close failed — <msg>` | `op: gh issue close failed for OP-72 · [Retry] · [Open log]` |
| Stale agent badge | (silent) | `op: OP-42 has no live agent · [Clear badge]` |
| tmux missing | `op: tmux not found at …` | `op: tmux not found at … · [Open settings]` |

### Files

- NEW `plugins/op-obsidian/src/actionableNotices.{ts,test.ts}` — helper + duration-heuristic unit tests
- NEW `plugins/op-obsidian/src/staleAgentBadges.{ts,test.ts}` — pure selector + 500ms-timeout tmux probe + tests
- NEW `plugins/op-obsidian/src/errorLog.ts` — vault-scratch error log writer/opener
- EDIT `plugins/op-obsidian/src/main.ts` — migrate sites; wire startup probe; centralize resolve / gh-close surfacing in `surfaceResolveResult` / `surfaceGhCloseError`
- EDIT `plugins/op-obsidian/src/resolve.ts` — drop UI Notices; surfacing is the caller's job now
- EDIT `plugins/op-obsidian/styles.css` — `.op-notice-action`
- BUMP `0.46.0 → 0.47.0` (minor — additive helper + new startup probe)

## Test plan

- [x] `npm run test` in `plugins/op-obsidian/` — 459 tests pass (added 8 new across two suites)
- [x] `npm run build` — clean (verified main.js fresher than manifest by `bump-version.mjs`)
- [x] Plugin reloaded in Agent-Vault at 0.47.0; `obsidian dev:console` clean after probe + synthetic Notice triggers
- [x] Synthetic `surfaceGhCloseError` rendered Notice with `[Retry]` + `[Open log]`; click on `[Open log]` navigated to `Projects/_scratch/op-last-error.md` (verified via `app.workspace.getActiveFile()?.path`)
- [x] Synthetic `surfaceResolveResult` rendered `op: TEST-RES resolved (1 task trashed) · [Open]`
- [x] Stale-agent probe ran clean — both `agent:`-bearing issues had live tmux windows so no Notices fired (correct behavior; manual kill-window test would surface the Notice)

### Pressure-test prompts for review

- **Click-handler lifecycle.** The action `<a>` elements live inside a `DocumentFragment` that becomes part of the Notice DOM; we capture the `Notice` reference in a closure to call `hide()` from the click handler. Is there a leak path if the user dismisses the Notice via the built-in close `×` before any action is clicked?
- **`[Retry]` re-entrancy.** The retry path calls `closeGithubIssue` directly. If the user spam-clicks `[Retry]`, we get N parallel `gh` invocations. The Notice doesn't currently disable the link after first click — is this acceptable, or should the retry be debounced?
- **Stale-probe race vs `op-work` mid-startup.** If `op-work` is mid-flight when `onLayoutReady` fires, an issue could have `agent:` written before its tmux window appears. The probe could fire a false-positive Notice during that microscopic window. Worth noting; mitigation would be an N-second post-probe delay, but the probe is already a one-shot and `[Clear badge]` is reversible — accepting it for now.
- **`[Open log]` privacy.** The error log can include the gh URL and repo path. Persisting these to a vault scratch note is more visible than the dev console — intentional (matches the `op-last-response.md` precedent), but flag if review thinks otherwise.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.46.0",
+  "version": "0.47.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.46.0",
+  "version": "0.47.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/actionableNotices.test.ts
+++ b/plugins/op-obsidian/src/actionableNotices.test.ts
@@ -1,0 +1,29 @@
+import { describe, it, expect, vi } from "vitest";
+
+vi.mock("obsidian", () => ({
+  Notice: class {
+    constructor(_msg: unknown, _duration?: number) {}
+    hide(): void {}
+  },
+}));
+
+import { durationForActions, type NoticeAction } from "./actionableNotices";
+
+const a: NoticeAction = { label: "Open", onClick: () => {} };
+
+describe("durationForActions", () => {
+  it("returns 0 (sticky) when actions are present", () => {
+    expect(durationForActions([a])).toBe(0);
+    expect(durationForActions([a, a])).toBe(0);
+  });
+
+  it("returns 5000 when no actions are present", () => {
+    expect(durationForActions([])).toBe(5000);
+  });
+
+  it("honours an explicit duration override", () => {
+    expect(durationForActions([], 1234)).toBe(1234);
+    expect(durationForActions([a], 7777)).toBe(7777);
+    expect(durationForActions([a], 0)).toBe(0);
+  });
+});
--- a/plugins/op-obsidian/src/actionableNotices.ts
+++ b/plugins/op-obsidian/src/actionableNotices.ts
@@ -1,0 +1,58 @@
+import { Notice } from "obsidian";
+
+/**
+ * One inline action link rendered after the Notice text. Clicking dismisses
+ * the Notice and runs `onClick`.
+ */
+export interface NoticeAction {
+  label: string;
+  onClick: () => void;
+}
+
+export interface ActionableNoticeOptions {
+  /** Leading prose. Plain text only. */
+  text: string;
+  /** Inline action links rendered after the text, joined by " · ". */
+  actions?: NoticeAction[];
+  /**
+   * Override duration (ms). Default: `0` (sticky) when `actions` is non-empty,
+   * else `5000` so non-actionable Notices keep current short-toast behavior.
+   */
+  duration?: number;
+}
+
+export function durationForActions(actions: NoticeAction[], override?: number): number {
+  if (typeof override === "number") return override;
+  return actions.length > 0 ? 0 : 5000;
+}
+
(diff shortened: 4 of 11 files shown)


earchibald:

1. **Click-handler lifecycle / leaks.** `actionableNotices.ts` captures the `Notice` in a closure and calls `notice.hide()` from each action's click handler. If the user dismisses the Notice via the built-in close `×` before clicking any action, the `<a>` listener stays attached to the (now-orphaned) fragment until GC. Is this a real leak or a non-issue given Obsidian's Notice teardown? If real, what's the right pattern — `AbortController.abort()` on `notice.hide()`?

2. **`[Retry]` re-entrancy.** `surfaceGhCloseError` shows `[Retry]` that calls `closeGithubIssue(repoPath, url, reason)` on click. The link doesn't disable after the first click; spam-clicking can fire N parallel `gh issue close` invocations. Is debouncing required, or is the retry path idempotent enough (second-call gracefully fails on already-closed) that this is fine?

3. **Stale-probe race.** `surfaceStaleAgentBadgesOnStartup()` runs once at `onLayoutReady`. If `op-work` is mid-flight (frontmatter has `agent:` set, tmux window not yet created), the probe could surface a false-positive `[Clear badge]` Notice. Mitigation options: (a) defer probe by N seconds, (b) probe twice with a delay, (c) accept the false positive since `[Clear badge]` is reversible. Which is right?

4. **`[Open log]` overwrite vs append.** `errorLog.ts` overwrites `op-last-error.md` on each error. If two `gh` failures fire in quick succession (rare but possible — bus listener + resolve hook double-fire), the second clobbers the first. Should this be append-only, or is overwrite-as-latest right (matches `op-last-response.md` precedent)?

5. **Probe failure semantics.** `probeLiveTmuxWindows` returns `{ ok: false }` on tmux missing/timeout, and the caller silently skips the surface step. Is silent-skip right, or should it emit a `[Open settings]` Notice nudging the user to fix tmux? (Could double-fire with the existing tmux-missing onload Notice.)

6. **Stale detection scope.** Currently the probe only checks the shared `op-agents` session. Per-window tmux sessions tracked in the orchestrator registry are excluded — is that correct, or should the probe also intersect with `tmuxSessionsForCleanup(reg)` like `agentSessionCleanup.ts` does?

7. **`[Start agent]` on op-new.** The action calls `doOpenAgent(res.entry, {})`. The newly-created issue is `open` (not `in-progress`); `doOpenAgent` → `openAgent` flips it to `in-progress` via `workIssue`. Is there any path where this short-circuits the user's expected flow (e.g. they intended `[Plan mode]` instead)? Should the Notice offer plan mode as a third action?

8. **Schema migration.** `runResolve` no longer emits Notices; callers must surface them. Anyone calling `runResolve` directly (not via `runResolveTracked`) would silently lose user feedback. Grep confirms only `runResolveTracked` is the caller — but flag if I missed an external entry point.

PR #171, opened 2026-04-25
OP-110: close orphan iTerm window when last agent resolves; bump to 0.48.0

earchibald:
## Summary

- When the last agent in an op-agents-N iTerm window resolves, `cleanupAgentSessions` killed the tmux window (and with it, the now-empty tmux session) but left the iTerm window alive with orphan PTYs. The next launch correctly reused the tmux session *name* but spawned a brand-new iTerm window — leaving visual clutter.
- Fix: add `closeWindow(windowId)` to the iTerm WebSocket driver and call it from cleanup when a window's last cell is freed. Best-effort (logged + swallowed); the WindowState is dropped from the registry either way.

## Test plan

- [x] `npx vitest run` — 463/463 pass (4 new tests in `agentSessionCleanup.test.ts`)
- [x] Bump to 0.48.0, build, sync to active vault, reload — `obsidian dev:console` clean
- [ ] Manual: launch two agents in op-agents-1, resolve both, launch a third — expect old iTerm window closed and a fresh one for op-agents-1

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.47.0",
+  "version": "0.48.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.47.0",
+  "version": "0.48.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/agentSessionCleanup.test.ts
+++ b/plugins/op-obsidian/src/agentSessionCleanup.test.ts
@@ -1,5 +1,6 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import {
+  cleanupAgentSessions,
   pruneRegistryForIssue,
   pruneRegistryMissingIssues,
   tmuxSessionsForCleanup,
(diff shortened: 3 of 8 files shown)


earchibald:
@copilot please pressure-test this PR. Specific concerns:

1. **CloseRequest semantics.** `force=true` is set to bypass iTerm's "process still running" confirmation. Are there iTerm builds (or a misconfigured user prefs) where `force` is ignored and the close silently no-ops? If so, our registry diverges from physical reality (window dropped from `reg.windows` but still on screen).

2. **Status enum coverage.** I treat `OK` and `NOT_FOUND` as success and throw on anything else. Per `api.proto:178-185`, `USER_DECLINED` is the third status. Could `USER_DECLINED` reach us with `force=true`? If so, throw vs swallow matters — caller catches and logs, so functionally fine, but it'd surface as a warning every cleanup. Worth pre-empting.

3. **Cleanup ordering vs concurrency.** `cleanupAgentSessions` runs sequentially: kill tmux → prune surfaces → close empty iTerm windows. If the user manually opens a new agent in the same iTerm window between phases (registry says "empty", iTerm has a fresh pane), we'd close their fresh pane. Race window is small (sub-second) but real. Should the empty-window check re-probe via `sessionExists` for the cells it nulled, instead of trusting the registry?

4. **Defensive orphan-surface sweep.** I added `for (const [issueId, ref] of Object.entries(args.reg.surfaces)) { if (ref.windowId === windowId) delete args.reg.surfaces[issueId]; }` after dropping the WindowState. Existing invariants (every assignSurface mirrors into sessionIds) say this should be a no-op. Is there any code path that breaks the invariant? `pruneDeadSessionSlots` is the only one I checked — it's symmetric.

5. **Test coverage.** I mocked `closeITermWindow` with a vi.fn. Did I miss the case where multiple windows are simultaneously empty (e.g., two windows each with one resolved agent — should both close in one cleanup pass)? My implementation iterates `Object.values` and the test only asserts the single-window case.

6. **iTerm window the user is *attached to*.** If the orphan iTerm window happened to be the user's currently-focused window, closing it shifts focus unpredictably. Is that surprising? (I suspect: yes, but the alternative — orphan stays — is what we're trying to fix. Acceptable tradeoff?)

7. **Backward compat.** `CleanupResult.closedWindows` is a new required field. Any callers that destructure the result and expect only the old shape? `grep CleanupResult` shows only `main.ts` and the test file.

Issue #154, opened 2026-04-25
Hotkey preset + op: pick & act modal (§3 of OP-149)


earchibald:
Tracked as op issue **OP-152**.

## Scope

- [ ] placeholder

PR #173, opened 2026-04-25
OP-152: hotkey preset + op: pick & act modal (§3 of OP-149)

earchibald:
## Summary

- Ships **§3 of OP-149** — a one-click default hotkey preset (10 bindings on `⌘⇧*` / `⌘⌥*`) and a new `op: pick & act` modal that combines fuzzy issue-picking with modifier-enter action dispatch (open / launch / plan / resolve / commit).
- Five new palette commands wired into `main.ts`: `op-pick-and-act`, `op-attach-current`, `op-resume-last` (stub awaiting §1), `op-next-issue`, `op-previous-issue` (numerically sorted, project-scoped, wrapping).
- Apply/revert routes through `app.hotkeyManager.setHotkeys`/`removeHotkeys` — direct `customKeys` mutation is silently discarded by Obsidian's getter-based snapshot. JSON-snippet fallback covers Obsidian API drift.
- 40 new unit tests (504/504 total). End-to-end smoke-tested via `obsidian eval`: apply, revert, modal open + fuzzy filter, next/previous round-trip.
- Bumped to **0.49.0** (minor).

Closes [OP-152](Projects/obsidian-projects/ISSUES/OP-152%20Hotkey%20preset%20%2B%20op%20pick%20%26%20act%20modal%20(%C2%A73%20of%20OP-149).md). Linked GitHub issue: #154.

## Test plan

- [x] `npm run build` clean
- [x] `npx vitest run` — 504/504 pass (40 new)
- [x] Smoke-tested in Agent-Vault: Apply preset → 10 bindings landed in `bakedHotkeys`; Revert → 10 cleared; `op: pick & act` modal opens with footer hint; fuzzy filter on "OP-152" surfaces the issue; `op-next-issue` rotates OP-152 → OP-153; `op-previous-issue` returns; `op-resume-last` stub fires Notice without throwing.
- [x] Console clean post-reload
- [ ] Adversarial Copilot review (requested below — touches public surface, new modal, hotkey API integration → all review-required per WORKFLOW.md)

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.48.0",
+  "version": "0.49.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.48.0",
+  "version": "0.49.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/hotkeyPreset.test.ts
+++ b/plugins/op-obsidian/src/hotkeyPreset.test.ts
@@ -1,0 +1,338 @@
+import { describe, it, expect, vi } from "vitest";
+import {
+  applyPreset,
+  canonicalKey,
+  defaultPreset,
+  revertPreset,
+  type AppLike,
+  type Hotkey,
+  type HotkeyEntry,
+} from "./hotkeyPreset";
+
+describe("defaultPreset", () => {
+  const preset = defaultPreset();
+
+  it("returns exactly 10 entries", () => {
+    expect(preset).toHaveLength(10);
+  });
+
+  it("returns the expected commands in spec-defined order", () => {
+    const ids = preset.map((e) => e.command);
+    expect(ids).toEqual([
+      "op-obsidian:op-open-sidebar",
+      "op-obsidian:op-pick-and-act",
+      "op-obsidian:op-resume-last",
+      "op-obsidian:op-attach-current",
+      "op-obsidian:op-open-agent",
+      "op-obsidian:op-resolve",
+      "op-obsidian:op-new",
+      "op-obsidian:op-append-commit",
+      "op-obsidian:op-next-issue",
+      "op-obsidian:op-previous-issue",
+    ]);
+  });
+
+  it("uses Mod+Shift or Mod+Alt only — Obsidian's least-reserved namespaces", () => {
+    for (const e of preset) {
+      const mods = new Set(e.hotkey.modifiers);
+      const isModShift = mods.size === 2 && mods.has("Mod") && mods.has("Shift");
+      const isModAlt = mods.size === 2 && mods.has("Mod") && mods.has("Alt");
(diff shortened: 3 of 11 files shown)


earchibald:
@copilot please review this PR with these specific pressure tests in mind:

1. **`hotkeyPreset.applyPreset` — collision detection edge cases.**
   - What if a user has a non-op command bound to *one* of the preset keys via `customKeys` (overriding a default that didn't match)? The conflict map's "first binding wins" branch — does it surface the *user's* override or the original default name? The user's override should win since it's what's actually active.
   - What if `customKeys[someId]` is `[]` (user explicitly cleared a default)? `effectiveBindings` returns `[]` so the slot looks free — verify this matches Obsidian's actual behavior under `bake()`. Worth testing live in case there's a stale `bakedHotkeys` entry.
   - What if two non-op commands conflict with the *same* preset binding? Only one is reported — is that the right UX, or should we surface the full list?

2. **`hotkeyPreset.revertPreset` — symmetry.**
   - The snapshot is captured from `customKeys` (a copy). If the user manually edited a hotkey *between* apply and revert, that edit is lost on revert — is that the intended behavior, or should we merge?
   - `removeHotkeys` clears overrides and restores defaults. If a preset binding *replaced* a user's prior override, revert restores the prior override; if it added a fresh override (no prior), revert clears it. Verify both paths.

3. **`pickAndActDispatch.decidePickAndActAction` — modifier ambiguity.**
   - macOS: `Cmd+Shift+Enter` — currently dispatches "launch" (Cmd wins). The default `op: resume last` binding *is* `⌘⇧↵` from the preset. Should pick & act's Cmd+Shift+Enter mean "launch" or "open the resume-last command"? Specifically, if the modal is open and the user hits the same shortcut they'd use to invoke `op-resume-last`, do we want pick-and-act to swallow it?
   - Cross-platform: on Linux, the spec describes `⌃↵` as "commit". My implementation requires `Ctrl+Alt+Enter` for commit on non-Mac (since plain Ctrl plays Cmd's role). Is that surprising?

4. **`PickAndActModal` — resolved-issue inclusion.**
   - The convention is "exact ID match only". A user typing `OP-7` (a partial that *is* an exact ID for OP-7) and `OP-72` (which is also an exact ID) — both work. But typing `OP-7` while OP-7 is resolved AND OP-72 is open: do we show both, only resolved OP-7, or only OP-72? Trace the filter and confirm.
   - Resolved issues sort to the bottom — but if the user typed an exact resolved ID, that's the only match they care about. Is the sort still useful, or should exact-ID matches jump to the top regardless of resolved status?

5. **`runRotateIssueCommand` — wraparound edge cases.**
   - Project with one open issue + active issue is that one: division by 1, idx=0, rotates to itself. Notice or no-op? Currently rotates to self silently.
   - Active issue is resolved (in `RESOLVED ISSUES/`): not in the `peers` filtered list, so `idx === -1` — falls through to "first/last open peer". Is that the right behavior, or should it Notice?

6. **`runPickAndActCommitFor` — git failures.**
   - Repo is detached-HEAD: `rev-parse --short=7 HEAD` succeeds, but the commit might already be reachable from another branch. Append happens regardless — intended?
   - `repo_path` is set but the directory was deleted: `execFile` rejects, the catch logs + Notices. No leak of partial commits to `commits:` since `appendCommit` is the last call. Verify.

7. **Settings `HotkeyPresetResultsModal` — Revert button after manual close.**
   - The Revert button calls `revertPreset(this.app, previousCustomKeys)` from the closure. If the user closes the modal without reverting, the snapshot is GC'd. That matches the spec ("session-scoped"), but verify there's no path where the snapshot leaks into persisted settings.
   - On the JSON-snippet fallback path, there's no Revert button rendered — confirm that's correct (nothing was actually written, so nothing to revert).

8. **Namespacing.** All 10 preset commands use `op-obsidian:` prefix. If the plugin is ever renamed, this hardcoding breaks. Worth pulling from `manifest.id` + the bare command id at runtime, or is the YAGNI win acceptable here?

PR #174, opened 2026-04-25
OP-167: persistent notification log + palette accessor

earchibald:
## Summary

- Notices dismiss on click, leaving no after-the-fact way to recover the text. This adds a vault-side persistent log at `Projects/_scratch/op-notifications.md` that captures every Notice the plugin shows.
- New palette command **`op: open notifications log`** opens the log in one click. The log is capped at 500 entries (newest-first) so it can't grow unbounded.
- `notify()` / `notifyAction()` wrappers replace `new Notice(...)` / `showActionableNotice(...)` across ~83 call sites. The `App` reference is registered once in `onload` so call-site signatures stay one-arg.
- Bump 0.48.0 → 0.49.0 (additive feature, no breakage).

Closes OP-167.

## Test plan

- [x] `npx vitest run` — 460 passing (15 new in `notificationLog.test.ts`).
- [x] `npx tsc --noEmit` — clean.
- [x] Smoke test in Agent-Vault: triggered failing `op-set-pr` and `op-append-commit` URIs; both Notices landed in the log newest-first; palette command opens the log; no console errors.
- [x] Verified concurrent appends serialize (race test in `notificationLog.test.ts`).

### DIFF
--- a/plugins/op-obsidian/src/editWorkflow.ts
+++ b/plugins/op-obsidian/src/editWorkflow.ts
@@ -1,4 +1,5 @@
-import { App, Notice, TFile } from "obsidian";
+import { App, TFile } from "obsidian";
+import { notify } from "./notificationLog";
 import type { OpSettings } from "./settings";
 import {
   AGENT_IDS,
@@ -37,14 +38,14 @@ export async function editWorkflow(
 ): Promise<EditWorkflowResult | undefined> {
   const trimmed = slug.trim();
   if (!trimmed) {
-    new Notice("op-edit-workflow: project slug is required");
+    notify("op-edit-workflow: project slug is required");
     return undefined;
   }
 
   const detection = detector.get() ?? (await detector.refresh());
   const installed = AGENT_IDS.filter((id) => detection?.[id]?.installed);
   if (installed.length === 0) {
-    new Notice("op: no supported agent binaries found on PATH");
+    notify("op: no supported agent binaries found on PATH");
     return undefined;
   }
 
@@ -55,7 +56,7 @@ export async function editWorkflow(
 
   const det = detection[agentId];
   if (!det.installed) {
-    new Notice(`op: ${agentId} binary not found on PATH (looking for "${profile.binary}")`);
+    notify(`op: ${agentId} binary not found on PATH (looking for "${profile.binary}")`);
     return undefined;
   }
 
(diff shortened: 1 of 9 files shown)


earchibald:
@copilot please review with these pressure tests in mind:

1. **Concurrent Notice storms.** Plugin startup fires several Notices in quick succession (tmux auto-detect, settings hydrate, agent detection probe). The write queue in `notificationLog.ts` serializes via a module-level promise chain — does it actually prevent read-modify-write races, or could two appends starting from the same snapshot of `existing` body still drop entries? Walk through the worst case.

2. **App not registered.** Tests register the App in `onload`. If a `notify(...)` somehow fires before `onload` or in a non-Obsidian test context, the log write is a silent no-op (the Notice itself still shows). Are there any code paths where this would silently lose data the user actually needs?

3. **Eviction edge cases.** `buildLog` does `[newLine, ...priorLines].slice(0, max)`. If `existing` body somehow contains lines starting with `- ` that aren't real entries (e.g. a stray bullet in the header text), the cap would mis-count. Confirm header contents can't pollute the entry-line filter.

4. **Notice text with control characters.** `formatLine` does `text.replace(/\\s+/g, " ")` to collapse whitespace. Does this handle null bytes, ANSI escapes from agent error output, or leading bullets that would break the markdown list? Specifically: if a Notice contains `\\n- not a real entry\\n`, the collapse turns it into ` - not a real entry `, which still parses as continuation text within a single bullet — confirm.

5. **Cap behavior at exactly MAX_LOG_ENTRIES.** Test exercises 550 → cap at 500. What about exactly 500 → 501 (oldest evicted) and 0 → 1 (initial). Is the boundary inclusive correct?

6. **`openNotificationLog` race.** If the user runs the palette command at the same instant a Notice fires, `writeOnce` and `openNotificationLog` could both `vault.create` the same path. Confirm one of them loses gracefully (Obsidian raises an exception that doesn't crash the plugin).

7. **Volume-on-startup performance.** ~50 startup Notices each trigger a read-modify-write of the same vault file. With a 500-entry cap this is bounded but each iteration reads + writes the full file. Acceptable on cold-start, or should we batch?

8. **Replacement diff symmetry.** Replaced `new Notice(` → `notify(` mechanically across ~83 sites. Did any call site pass a non-string first argument (DocumentFragment, etc.) that would now misbehave? Spot-check `actionableNotices.ts` is the only intentional `new Notice(frag, ...)` left.

9. **Settings tab side effects.** Several Notices in `settings.ts` fire during validation that can repeat rapidly (user typing in a JSON field). Does spamming the log here matter, or is the cap sufficient?

10. **Cleanup symmetry on plugin unload.** Module-level `registeredApp` and `writeQueue` are not cleared on `onunload`. If the plugin reloads, the chain still references the old App. Should `registerApp` be called in `onload` (already is) — confirm a stale `writeQueue` from a prior session can't fire after `onunload`.

PR #175, opened 2026-04-25
OP-150: op:resume-last + Last-touched sidebar chip + visibility-gated tmux liveness probe

earchibald:
## Summary

Implements §1 of OP-149 (resumability):

- New palette command **`op: resume last`** opens the most-recently-touched issue note and, on macOS, re-attaches to its agent via the existing `revealAgentSession` path when the tmux window is still alive.
- New persistent **recency log** in `data.json` (`recent: RecencyEntry[]`, capped at 25, dedup-by-id, `Array.isArray` self-healing read). Records a touch on every dispatch point that opens or attaches an issue: palette `op-work` / `op-open-agent`-and-variants, their URI handlers, the `op-work` CLI handler, `op-attach-current`, and the sidebar row click.
- New **Last-touched sidebar chip** above the tabs (hidden when empty; click triggers `op: resume last`).
- New **visibility-gated tmux liveness probe** in the sidebar: 5 s `setInterval`, started in `onOpen`, cleared in `onClose`, gated on `process.platform === "darwin"`. Issues whose `agent:` field claims a session tmux no longer has render with `op-sidebar__agent--stale / is-stale` — kept clickable, visibly muted.

## Key design call: reuse `probeLiveTmuxWindows`, no new `tmuxLiveness.ts`

The spec sketch called for a new `tmuxLiveness.ts` module with an `execFile` wrapper, 500 ms timeout, and exact-match parse against `tmuxWindowName`. Those invariants already live in `staleAgentBadges.ts::probeLiveTmuxWindows` (used by the startup stale-badge probe) and the matching pure `selectStaleAgentBadges`. Adding a parallel module would have been a second indirection to keep in sync. Funneling the sidebar's periodic probe through the same helper preserves the spec's risk mitigation ("all probes go through one module so the timeout can't be bypassed").

The macOS gate lives at the call site (sidebar's `startTmuxProbe`), not inside `probeLiveTmuxWindows` — on Linux the probe would otherwise burn 500 ms × 12 ticks/min for nothing.

## Test plan

- [x] `npx vitest run` — full suite passes (513 tests)
- [x] `recencyLog.test.ts` — read/write/cap/idempotent-move-to-head, array-guard against `null`, `{}`, `[1,2]`
- [x] `sidebarView.test.ts` — visibility-timer lifecycle: starts on `onOpen`, stops on `onClose`, no double-start, no probe on non-darwin or without hooks
- [x] Smoke test against Agent-Vault:
  - [x] `recordRecency` dedupes and caps in live `data.json`
  - [x] `op: resume last` opens the head issue note (verified post-`await` because the eval bash returned before the `openFile` resolved)
  - [x] Sidebar chip renders `Last touched: OP-150 · claude attached · 1m ago` and re-renders after `recordRecency`
  - [x] Forcing `liveTmuxWindows = new Set(["OP-150"])` demotes the OP-123 badge to `is-stale` while leaving OP-150 fresh
  - [x] Detaching the sidebar leaf flips `isProbeRunning()` from `true` → `false` with no console errors
- [x] No `obsidian dev:console` errors after a full open → recordRecency → resume → close cycle

## Risks (delta from issue body)

- The "all probes through one module" mitigation is preserved by reusing `probeLiveTmuxWindows`. No new wrapper to keep in sync.
- The `recent:` field is sanitized through `sanitizeRecency` on every read, so hand-corrupting `data.json` (typing garbage, replacing with `null`/`{}`) leaves the rest of settings intact.
- `OpSidebarHooks` is optional in the `OpSidebarView` constructor — legacy callers (existing sidebar tests) keep working unchanged; only `main.ts` wires the full hook set.

## Pressure-test prompts for adversarial review

1. **Concurrent `op-work` calls**: if `recordRecency` is invoked twice rapidly (palette + URI for the same id), do we end up with two entries in `recent:`, or is the dedup atomic? Is there a race between two `await this.saveSettings()` calls?
2. **Cap-25 invariant** when an existing entry is moved to head: confirm the result is still ≤ 25 in `appendRecency` regardless of input length.
3. **`op-open-agent` failure path**: when `openAgent` returns `undefined` (cancelled, no agent installed), do we accidentally still record the issue? (Should not — both call sites check `res` truthiness first.)
4. **Sidebar probe on a darwin host where tmux is stopped mid-session**: does `probeLiveTmuxWindows` reliably return `{ok:false}` or do we ever throw out of the timer callback and crash the next tick? See the try/catch in `tickTmuxProbe`.
5. **Schema migration**: existing users have `data.json` without a `recent:` field. Verify `mergeSettings` produces `recent: []` (does it, given my `sanitizeRecency(undefined) === []`?).
6. **`refresh()` is `public` but not used internally** — is exposing it a footgun, or is it the right escape-hatch for tests?
7. **CSS specificity**: `op-sidebar__agent--stale` overrides `op-sidebar__agent--claude`'s background — confirm the cascade order in `styles.css` produces the muted look across all three agents (claude/gemini/copilot).
8. **`op:resume-last` when the head issue's note has been moved to `RESOLVED ISSUES/`**: my `byId` lookup hits the resolved entry; do I open it correctly, or does the resolved fork need different handling?

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.50.1",
+  "version": "0.51.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.50.1",
+  "version": "0.51.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/main.ts
+++ b/plugins/op-obsidian/src/main.ts
@@ -1091,21 +1091,45 @@ export default class OpPlugin extends Plugin {
    * entry from the recency log, opens its issue note, and (on darwin only)
    * re-attaches to the agent's tmux window if it's still alive.
    *
+   * Stale entries (issue files deleted from the vault) are pruned eagerly in a
+   * single pass so that one invocation always lands on the first surviving entry
+   * rather than requiring repeated invocations to clear tombstones one at a time.
+   *
    * On non-macOS platforms, falls through to "open the note" — the recency
    * log still works as a navigation aid even when terminal launch isn't
    * supported.
    */
   private async runResumeLastCommand(): Promise<void> {
-    const head = mostRecent(this.settings.recent);
+    // Prune all leading tombstones in one O(n) pass so a single invocation
+    // always reaches the first surviving entry (§8 adversarial: multiple
+    // deleted heads — would otherwise require one invocation per stale entry).
+    let cursor = 0;
+    while (cursor < this.settings.recent.length) {
+      const candidate = this.settings.recent[cursor];
+      const e = this.store.byId(candidate.id);
+      if (e && e.type === "issue") break;
+      cursor++;
+    }
+    const staleCount = cursor;
+    if (staleCount > 0) {
+      const staleIds = this.settings.recent.slice(0, staleCount).map((e) => e.id);
+      this.settings.recent = this.settings.recent.slice(staleCount);
+      await this.saveSettings();
+      if (staleCount === 1) {
+        new Notice(`op: ${staleIds[0]} is no longer in the vault — cleared from recency log.`);
+      } else {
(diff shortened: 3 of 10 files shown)


earchibald:
@copilot please run an adversarial review focused on the pressure-test prompts in the PR description (sections 1–8). Particular concerns:

- §1 race between two `recordRecency` writes — does `await saveSettings()` linearize, or can two concurrent invocations clobber each other's `recent:` array?
- §3 silent recording on cancelled `openAgent` — verify both palette and URI paths.
- §4 `tickTmuxProbe`'s try/catch as the timer's last line of defense — confirm it can't blow up the interval.
- §8 resolved-issue head: open or skip?

The key design choice is reusing `probeLiveTmuxWindows` instead of building a parallel `tmuxLiveness.ts`. That deviates from the issue body's literal Files-likely-touched list — please flag if you think the parallel module would have been the better call given the spec's "funnel all probes through one module" risk mitigation.

PR #176, opened 2026-04-25
OP-123: op-set-link / op-remove-link palette commands

earchibald:
## Summary
- Add `op: set issue link` and `op: remove issue link` to the command palette — three-step modal flow (source → relation → target) wired to the existing `applyLink` / `removeLink` handlers.
- New `RelationPickerModal` (FuzzySuggestModal over `RELATION_NAMES`); new `listLinkedTargets` helper enumerates currently-linked issues so `op-remove-link`'s target picker is constrained to actual links.
- Bumps op-obsidian to **0.50.0** (minor — additive feature, no schema or surface removals).

## Implementation notes
- No new URI handler; CLI/URI surface for `op-set-link` / `op-remove-link` was already shipped in OP-113.
- Empty-relation edge case for `op-remove-link`: show a Notice and stop, no fall-through to a free-text picker.
- Many-to-one re-parent cleanup is inherited from `applyLink` unchanged.
- Tests: 6 new vitest cases for `listLinkedTargets` covering list/scalar/empty/dangling-target/unknown-source/unknown-relation. Full suite passes (519 tests).

## Test plan
- [x] `npx vitest run` — all 519 tests pass.
- [x] `obsidian plugin:reload id=op-obsidian` — both new commands appear in `app.commands.commands` (`op-obsidian:op-set-link`, `op-obsidian:op-remove-link`).
- [x] Console clean after running each command (no errors).
- [x] Both commands open the FindIssueModal as their first step.
- [ ] Manual: pick source → relation → target, confirm Notice and verify two-sided frontmatter write.
- [ ] Manual: `op-remove-link` against an issue with no `parent` — confirm "no parent links to remove" Notice.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.49.0",
+  "version": "0.50.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.50.0",
+  "version": "0.50.1",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.50.0",
+  "version": "0.50.1",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.49.0",
+  "version": "0.50.0",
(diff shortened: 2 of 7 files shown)


earchibald:
@copilot — adversarial review please. Specific pressure-test prompts:

1. **Three-step modal lifecycle.** If the user dismisses the relation picker (Esc) after picking the source, do we leak the source pick or accidentally re-open something? Walk through `runSetLinkCommand` → `pickSetLinkTarget` and confirm there's no zombie callback if any modal in the chain is closed without a choice.

2. **`op-remove-link` empty-state vs. drift.** `listLinkedTargets` silently drops link entries whose target id no longer resolves (link drift). For `op-remove-link`, that means a relation with two links — one live, one dangling — only shows one in the picker. Is that the right call, or should we surface the dangling entry too so the user can clean it up? (`op-link-check-repair` exists, but it's a separate flow.)

3. **Many-to-one re-parent via palette.** Picking `parent` and a new target on an issue that already has a parent triggers `applyLink`'s cleanup of the old holder's `children:` list. The palette handler shows the cleaned ids in a Notice — does anything about doing this from the GUI surprise the user vs. the CLI verb? Confirm there's no missing confirmation step.

4. **Self-link guard surfacing.** `validateLinkArgs` throws "Cannot link issue to itself" — `doApplyLink` catches and shows a Notice. Verify the error path: is the message readable, and does the modal close cleanly after the Notice?

5. **Symmetric relations (`related_to ↔ related_to`).** When the source has `related_to: [OP-2]` and we run `op-remove-link related_to`, the picker offers OP-2 — fine. But on `op-set-link related_to OP-1 → OP-2` from the palette, the bidirectional write should land on both ends. Eyeball the chain to confirm we're not double-appending or skipping the inverse.

6. **`RelationPickerModal` placeholder.** Currently shows e.g. "Pick relation for OP-123". Does fuzzy-matching work over the relation names (parent, children, depends_on, depended_on_by, related_to)? Any naming collisions a typo'd query would accidentally match against?

7. **Test coverage.** `links.test.ts` covers `listLinkedTargets` but not the new `runSetLinkCommand` / `runRemoveLinkCommand` orchestration. Is that gap acceptable given those methods are pure UI glue (modals + Notice + delegating to already-tested `applyLink` / `removeLink`)?

PR #177, opened 2026-04-25
OP-139: add 2x3 layout (6 cells); new default

earchibald:
## Summary

- Adds a new `2x3` orchestrator layout (2 rows × 3 cols = 6 cells), filling the gap between `2+3` (5) and `2x3+2` (8).
- Switches `DEFAULT_SETTINGS.orchestrator.preferred` from `2x2` (4 panes) to `2x3` (6 panes). Existing users with a saved value keep it; only fresh installs get the new default.
- `pickLayout(6)` now returns `2x3` instead of `2x3+2` (smallest fitting layout).
- README updated to list the full layout menu and call out the new default.
- Bumps lockstep to `0.51.0` (minor — additive layout + new default value is user-visible new behavior).

## Test plan

- [x] `npm test` in `plugins/op-obsidian/` — 546 passed (was 534, +12 from new assertions).
- [x] Built `main.js` synced into Agent-Vault and reloaded.
- [x] Smoke-tested Settings tab: dropdown lists `1, 1x2, 3, 2x2, 2+3, 2x3, 2x3+2, 3x3` (the new option appears between `2+3` and `2x3+2`).
- [x] No console errors after opening the Settings tab.
- [ ] Adversarial Copilot review (requested below).

### DIFF
--- a/plugins/op-obsidian/README.md
+++ b/plugins/op-obsidian/README.md
@@ -102,7 +102,7 @@ Per-project repo paths (slug → absolute path). Overridden by a `repo_path:` in
 
 ### iTerm layout orchestrator
 
-Optional. When enabled, op arranges agent panes in the current iTerm window using the chosen layout (`1x1`, `2x1`, `2x2`, `3x2`, …). Overflow spills to a new iTerm window with a fresh tmux session. macOS + iTerm only.
+Optional. When enabled, op arranges agent panes in the current iTerm window using the chosen layout (`1`, `1x2`, `3`, `2x2`, `2+3`, `2x3`, `2x3+2`, `3x3`). Default is `2x3` (6 panes). Overflow spills to a new iTerm window with a fresh tmux session. macOS + iTerm only.
 
 ### Sidebar view
 
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.50.1",
+  "version": "0.51.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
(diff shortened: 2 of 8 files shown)


earchibald:
@copilot adversarial review please. Specific pressure-tests:

1. **Default-change blast radius.** `DEFAULT_SETTINGS.orchestrator.preferred` flipped from `2x2` to `2x3`. Verify `mergeSettings()` in `settingsPure.ts` still preserves a user's explicitly-saved value (e.g. someone who chose `2x2` should keep `2x2` after upgrade). Look for any path where a missing or partially-loaded `orchestrator` block could clobber the saved value with the new default.

2. **Layout schema invariants.** The 2x3 layout I added has `cells: 6`, `splits.length: 5`, and each split's `from` references an already-created cell index. Verify the `layouts.test.ts` invariants still hold for it (split count = cells - 1; every `from < cellsBefore`). Also check the visual ordering claim — the splits produce TL, BL, TM, TR, BM, BR (creation order), which matches the 3x3 pattern. Anything off in the geometry?

3. **`pickLayout(6)` drift.** Previously 6 agents → `2x3+2` (8 cells, two empty slots). Now → `2x3` (exact fit). Check `engine.ts` for any caller that relied on the old over-allocation behavior. Also: `pickLayout(7)` now still returns `2x3+2` because the smallest-fitting filter still excludes `2x3` for count=7 — verify the test I added (`pickLayout(7) === "2x3+2"`) actually proves that.

4. **`maxLayoutForCeiling` change.** With `maxRows=2, maxCols=3`, `maxLayoutForCeiling` now returns `2x3` (6 cells) instead of `2+3` (5 cells). No existing test covers that combination — is there any caller that relies on the old behavior?

5. **Settings dropdown UI.** The dropdown iterates `LAYOUT_IDS` directly. Smoke-tested manually and `2x3` shows between `2+3` and `2x3+2`. Anything else that hard-codes the layout list (palette commands, sidebar, status messages)?

6. **Registry / persistence compatibility.** `RegistryData.windows[*].layoutId` is typed `LayoutId`. Existing saved-state with `"2x2"` is still in the union. Verify a saved registry pointing at `2x2` from before the upgrade still loads cleanly through `mergeRegistry()`.

7. **Version bump.** Lockstep across all three JSON files; minor was the right level (additive + new user-visible default).

PR #179, opened 2026-04-25
OP-157: sidebar keyboard nav + density preference (§6 of OP-149)

earchibald:
## Summary

- New `view.density: 'comfortable' | 'compact'` setting; `compact` tightens vertical row padding by 4px and hides the project chip when every rendered row shares one project (`shouldShowProjectChip` helper, unit-tested).
- Keyboard nav inside the sidebar leaf: `j`/`k`/`ArrowDown`/`ArrowUp` cycle `.is-selected`, `Enter` opens, `Cmd/Ctrl+Enter` launches an agent, `r` opens a confirmation modal that calls a new `resolveIssue` hook (wired to `runResolveCommand`). Listener is scoped to the sidebar's `contentEl` and skips letter keys when the filter input owns focus.
- The Last-touched header chip from §6 already shipped in OP-150 (c3464a4), so no chip plumbing was needed in this PR.
- Bumped op-obsidian 0.51.0 → 0.52.0 (additive feature; two new public surfaces).

Spec: `docs/specs/OP-149-feel-great-ux-design.md` §6.

## Test plan

- [x] Unit tests added: `view.density` accept/reject in `settings.test.ts`; `shouldShowProjectChip` cases in `sidebarView.test.ts`. Full suite: 570/570 pass.
- [x] Settings UI smoke: `app.setting.openTabById("op-obsidian")` shows the new "Sidebar density" dropdown.
- [x] Live keyboard smoke in Agent-Vault: dispatched `KeyboardEvent`s for `j`, `k`, `ArrowDown`, `ArrowUp` cycle the `.is-selected` index across the rendered list (verified across 30 issues).
- [x] `r` opens the resolve confirmation modal (title `Resolve OP-NNN?`, Cancel + Resolve buttons); Cancel dismisses without mutating the vault.
- [x] Density toggle: `compact` + filter to a single project drops the project chips (21 rows, 0 chips); `comfortable` restores them; class `op-sidebar--density-compact` toggles on the sidebar root.
- [x] Build green; no console errors after a full reload.
- [ ] User verification of the keyboard ergonomics in their day-to-day flow.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.51.1",
+  "version": "0.52.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.51.1",
+  "version": "0.52.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/main.ts
+++ b/plugins/op-obsidian/src/main.ts
@@ -334,6 +334,7 @@ export default class OpPlugin extends Plugin {
             tmuxSessions: () => tmuxSessionsForCleanup(this.settings.orchestratorState),
             recordRecency: (id) => this.recordRecency(id),
             executeResumeLast: () => void this.runResumeLastCommand(),
+            resolveIssue: (entry) => this.runResolveCommand({ path: entry.path }),
           },
         ),
     );
--- a/plugins/op-obsidian/src/settings.test.ts
+++ b/plugins/op-obsidian/src/settings.test.ts
@@ -120,6 +120,18 @@ describe("mergeSettings", () => {
     ).toBe(DEFAULT_SETTINGS.view.defaultTab);
   });
 
+  it("view.density accepts the two literals; rejects anything else", () => {
+    expect(DEFAULT_SETTINGS.view.density).toBe("comfortable");
+    expect(mergeSettings({ view: { density: "compact" } }).view.density).toBe("compact");
+    expect(mergeSettings({ view: { density: "comfortable" } }).view.density).toBe("comfortable");
+    expect(
+      mergeSettings({ view: { density: "tight" as unknown as "compact" } }).view.density,
+    ).toBe(DEFAULT_SETTINGS.view.density);
+    expect(
+      mergeSettings({ view: { density: 1 as unknown as "compact" } }).view.density,
+    ).toBe(DEFAULT_SETTINGS.view.density);
+  });
+
   it("only accepts terminal === 'Terminal' | 'iTerm'", () => {
     expect(mergeSettings({ terminal: "iTerm" }).terminal).toBe("iTerm");
     expect(mergeSettings({ terminal: "Terminal" }).terminal).toBe("Terminal");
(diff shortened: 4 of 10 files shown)


earchibald:
@copilot — adversarial review please. Concrete pressure-test prompts:

1. **Keydown handler scope leakage.** The handler is attached to `this.contentEl` and the leaf gets `tabindex=0` + auto-focus. If the user clicks into another pane (editor, file explorer, search) and back, does the sidebar steal `j`/`k`/`r`/`Enter` from the legitimate target? Verify the handler only fires when focus is inside the sidebar's contentEl tree, and that detached/closed sidebars don't leak listeners.

2. **`r` resolve race.** What happens if the user holds `r` (auto-repeat) — does each keystroke open a fresh modal, stacking them? What if the user spams `r` while a previous resolve is still in-flight (the `op-resolve` operation is async, moves files, runs `gh issue close`)? Confirm there's no double-resolve risk and no orphaned modals.

3. **Selected index drift across store events.** Issues are sorted by id; if a fresh issue arrives via `issue:created` with an id that sorts before the current selection (e.g. selected `OP-100`, then `OP-001` lands), the index now points at a different row. Is the row identity preserved or does the highlight silently jump? If it jumps, is that surprising for the user mid-keypress?

4. **`Cmd+Enter` collision.** Obsidian's default `Cmd+Enter` is "Toggle checkbox status" / "Open link in new pane" depending on context. Verify `event.preventDefault()` actually wins inside an item-view leaf (not just markdown editors), and that the launch doesn't double-fire if both handlers run.

5. **Density chip suppression edge cases.** The single-project check uses `entries[0].project` as the reference. What if the rendered list is `[{project: "a"}]` then becomes `[{project: "a"}, {project: "b"}]` mid-render via a store event — does the chip flicker on/off? Is `shouldShowProjectChip` called in a hot path that needs memoization?

6. **Filter-input letter-key gating.** The handler skips `j`/`k`/`r` when `target === filterInput` but still honors Arrow keys, Enter, Cmd+Enter. Does the keyboard router correctly distinguish "focus inside the leaf but not in the input" vs "focus on the input"? What about a future child input element (the resolve modal's buttons, for instance) — does the gate hold?

7. **CSS specificity on `.op-sidebar__row.is-selected`.** The selected-row style uses `box-shadow: inset 0 0 0 1px var(--interactive-accent)`. Does this conflict with the existing hover state on `.op-sidebar__link:hover`, or with the agent badge backgrounds? Test with the resolved tab and with the filter input focused.

8. **Mobile / touch.** The `r` shortcut and `tabindex=0` auto-focus may have side effects on Obsidian Mobile (no physical keyboard but the focus ring still applies). Is there a regression risk for tablet users?

9. **Backwards-compat on legacy settings.** Loading a `data.json` from v0.51 (no `view.density` field) — does `mergeSettings` cleanly default to `"comfortable"`? Confirm by hand-deleting the field and reloading.

10. **`scrollIntoView` jank.** With long lists (50+ issues), each `j`/`k` keystroke fires `scrollIntoView({block: "nearest"})`. On low-end hardware does this cause visible flicker, or does the implicit smooth-scroll-disabled behavior keep it instant? Should we throttle / RAF-batch?

earchibald:
Copilot — thanks for the four review comments and the three follow-up fixes in ea94c8c.

I rebased the branch onto origin/main (OP-141 + OP-139 had landed) and accidentally force-pushed over your fix commit. Recovered ea94c8c via SHA before GC, then re-applied your three fixes by hand on top of my refactored `decideKeyAction` router:

1. `r` auto-repeat modal stacking guard — `resolveModalOpen` flag + `onDismiss` callback wired into `OpResolveConfirmModal.onClose`.
2. Selection identity preservation by id — `selectedId` snapshot + `findIndex` restore in `render()`, with the original index-clamp fallback when the selection leaves the rendered list.
3. `Cmd/Ctrl+Enter` `stopPropagation()` in the launch path.

Plus your four unit tests (modal onDismiss + no-onDismiss + selection-identity preserved + clamp-fallback). All 585 vitest tests pass; live smoke confirms spamming `r` opens exactly one modal and that the next `r` after dismissal works.

Re your other four review comments addressed in d44ebc7: `tabindex=0` comment fixed, `:focus-visible` ring restored, two-context docstring fixed, and the keyboard-binding map lifted into a pure `decideKeyAction(ev, opts)` helper with 7 unit tests.

Latest sha: `ca29210`.

PR #180, opened 2026-04-25
OP-142: adopt op-set-section across op:issue, op:resolve, plan-mode persistence

earchibald:
## Summary

- Migrate every Plan/Notes/Summary write off direct `Edit`/`op-set-scope mode=body` and onto the dedicated `op-set-section` verb shipped in OP-125. The strongest motivator: plan-mode's persistence path used `op-set-scope mode=body`, which clobbers every section outside `## Scope` — including any work-in-progress Plan, Tasks, Notes, or Summary content.
- Skill (`SKILL.md`) gets a new `op-set-section` row in the commands table; `verb:work` Plan/Notes-append steps and `verb:resolve` Summary step are rewritten to call the verb explicitly. `cli-gotchas.md` gains a "Body sections without a verb" note that names `op-set-section` as preferred and the read-then-rewrite pattern as the plugin-disabled fallback only.
- `agentProfiles.ts` plan-mode preamble + agent definition switch the sanctioned mutation from `op-set-scope` to `op-set-section name=Plan`. Evaluate-mode is intentionally left alone: it persists `## Initial Evaluation`, which is not in the `Plan|Notes|Summary` enum.
- Lockstep version bump `0.51.0 → 0.52.0` (additive: sanctions a new mutation surface in plan mode).

## Test plan

- [x] `npm run build` (via `node scripts/bump-version.mjs minor`) green.
- [x] Targeted vitest: `npx vitest run agentProfiles uriHandlers cliHandlers setSection` → `PASS (99) FAIL (0)`.
- [x] Synced rebuilt `main.js` + `manifest.json` into the active vault, reloaded `op-obsidian` (`obsidian plugin:reload id=op-obsidian`).
- [x] Smoke test against this very issue: `obsidian op-set-section issue=OP-142 name=Notes content="…" append=true` extended `## Notes` and left `## Plan`, `## Scope`, `## Tasks`, `## Initial Evaluation`, `## Summary` untouched.

## Adversarial review pressure-test prompts (for `@copilot`)

- The plan-mode preamble drops the `op-set-scope mode=body` escape hatch entirely. Is the new `op-set-section name=Plan`-only sanction strictly weaker for any legitimate plan-mode workflow that needs to write multiple H2 sections? If so, what's the right fallback — emit multiple `op-set-section` calls in sequence, or extend the enum?
- `op-set-section` rejects payloads containing `## ` H2s. The Notes-append flow uses `### <ID>.<N>` sub-headings — confirm no edge case where an agent might inadvertently include `##` (e.g. nested in a fenced code block) and have the verb reject the call. What's the failure mode if it does?
- The Notes-append idempotency story now lives at the agent layer ("read first, branch on whether the heading already exists"). Is that adequately spelled out in the SKILL.md update, or could two parallel agents each call `append=true` and produce duplicate blocks?
- Evaluate-mode still uses `op-set-scope mode=body` to write `## Initial Evaluation` (because the enum doesn't include that name). Is there a race where evaluate-mode's body-clobber overwrites an in-progress Plan that was just persisted via `op-set-section`?
- Version bump `0.51.0 → 0.52.0` collides with a version already deployed in at least one local Obsidian vault sync. Confirm the canonical source-of-truth is the bumped `manifest.json` in this PR, not the in-vault artifact (which is from an unmerged worktree).


earchibald:
@copilot Please pressure-test this change against the prompts below. This PR migrates the op-set-section verb from "available but unused" to "primary mutation surface for Plan/Notes/Summary," and removes the `op-set-scope mode=body` escape hatch from plan-mode entirely.

1. **Plan-mode capability narrowing.** The plan-mode preamble used to sanction either `op-set-scope` (rewrite `## Scope`) or `op-set-scope mode=body` (rewrite the whole post-title body). Now only `op-set-section name=Plan` is sanctioned. Is this strictly weaker for any legitimate plan-mode workflow that needed to span multiple H2s in one persist? If yes, what's the right fix — emit a sequence of `op-set-section` calls, or extend the verb's enum?
2. **Payload H2 rejection.** `op-set-section` rejects payloads containing `## ` H2s (would terminate the section). The Notes-append flow encourages `### <ID>.<N>` sub-headings — but agent-generated content could plausibly nest a `##` inside a fenced code block. Confirm whether `setSectionPure.ts` strips/validates inside fences or treats every `## ` line as terminating, and whether the failure surface is loud enough.
3. **Concurrent Notes appends.** With `append=true` and the read-first-then-update-in-place idempotency rule shifted to the agent layer (per the new SKILL.md text), is there a race where two agents both detect "block doesn't exist yet" and both append, producing duplicates? If yes, should the verb itself dedupe by `### <heading>` line?
4. **Evaluate-mode untouched.** Evaluate-mode still uses `op-set-scope mode=body` to write `## Initial Evaluation`. Is there a sequence where evaluate-mode runs after plan-mode has already persisted `## Plan` via `op-set-section`, and the evaluate-mode body-clobber wipes the plan? If yes, this PR has shipped a regression risk.
5. **Documentation/code drift in cli-gotchas.md.** The new "Body sections without a verb" section recommends a read-then-rewrite fallback when the plugin is disabled. Does that fallback faithfully replicate `op-set-section`'s payload constraints (name enum, no H2 in content), or could an agent following the fallback produce a malformed body that `op-set-section` would have rejected?
6. **Version collision.** Lockstep bumped `0.51.0 → 0.52.0`. Confirm there's no other open PR or unmerged branch already at `0.52.0` — and if there is, that the squash-merge order makes one of them rebase and bump again rather than silently overwriting.

PR #181, opened 2026-04-25
OP-143: sidebar hover preview shows last N lines of agent's tmux pane

earchibald:
## Summary

- New sidebar interaction: hovering an agent badge (e.g. `claude`) shows a small popover with the last N lines of that agent's tmux pane after a configurable delay. Suppressed silently when no live tmux window exists.
- Three new `view.*` settings (`agentHoverPreview` toggle, `agentHoverLines` / `agentHoverDelayMs` numbers) under "Sidebar view" — defaults: on, 30 lines, 400 ms.
- Pure refactor inside: `findAgentTmuxLocation` extracted from `revealAgentSession.ts` into `agentTmuxLocation.ts` so reveal-on-click and the hover preview share the shared-session + per-iTerm-window orchestrator session lookup.

Implements OP-143. Bumps op-obsidian to **0.53.0** (additive user-facing behavior).

## Test plan

- [x] `npx vitest run` — 559 / 559 pass (new `tmuxCapture.test.ts` + 4 new cases in `settings.test.ts`).
- [x] `npx tsc --noEmit` clean.
- [x] Bump+build via `scripts/bump-version.mjs` (lockstep across `manifest.json`, `package.json`, `plugin.json`); main.js fresher than manifest.
- [x] Synced into Agent-Vault; `obsidian plugin:reload id=op-obsidian` clean.
- [x] DOM-asserted the three new settings rows via `app.setting.openTabById("op-obsidian")`.
- [x] Hover smoke: live-tmux badge → popover renders the last 30 lines after the delay; mouseleave tears it down; rapid hover+leave cancels the pending capture; disabling `agentHoverPreview` suppresses it entirely.
- [x] `obsidian dev:console` — no errors after exercising the flow.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.53.0",
+  "version": "0.54.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.53.0",
+  "version": "0.54.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/agentTmuxLocation.ts
+++ b/plugins/op-obsidian/src/agentTmuxLocation.ts
@@ -1,0 +1,63 @@
+import { execFile } from "child_process";
+import { promisify } from "util";
+
+import type { OpSettings } from "./settings";
+import { SHARED_TMUX_SESSION, tmuxWindowName } from "./terminalLaunch";
+
+const pExecFile = promisify(execFile);
+
+export interface AgentTmuxLocation {
+  session: string;
+  window: string;
+}
+
+/**
+ * Resolve an issue id to the (session, window) pair of its live tmux pane,
+ * or null if no candidate session contains the expected window. Iterates
+ * the shared session plus every per-iTerm-window orchestrator session,
+ * returning the first match.
+ *
+ * Pure read — never mutates tmux state. Suitable for hover-preview captures
+ * as well as `revealAgentSession`'s select-window flow.
+ */
+export async function findAgentTmuxLocation(
+  settings: OpSettings,
+  issueId: string,
+): Promise<AgentTmuxLocation | null> {
+  const surface = settings.orchestratorState?.surfaces?.[issueId];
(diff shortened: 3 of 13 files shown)


earchibald:
@copilot — adversarial review please. This adds a new sidebar interaction (hover an agent badge → tmux pane preview popover) and three new `view.*` settings. Touches `sidebarView.ts`, `settingsPure.ts`, `settings.ts`, `main.ts`, plus a refactor extract from `revealAgentSession.ts` into a new `agentTmuxLocation.ts` module.

Please pressure-test:

1. **Stale capture races.** `attachHoverPreview` debounces `agentHoverDelayMs` then calls `capturePreview(entry)`. The `hoverSeq` is bumped on both mouseenter and mouseleave. Is there any sequence where a stale tmux capture lands in the popover after the user has moved on or hovered a different badge? Consider rapid hover-A → hover-B (bumps seq twice) → A's capture resolves last. Are we guarded?

2. **Popover lifecycle leaks.** `hoverEl`, `hoverTimer`, and `hoverDocClickHandler` are instance fields. `teardownHoverPreview` is invoked from `onClose` and at the start of every `render` (which rebuilds the list). `setSelectedIndex` does not call `render` — does the keyboard-nav path (j/k) avoid orphaning the popover? Is there a path where the leaf detaches without calling `onClose`?

3. **Click-outside dismissal.** Document-level click handler attached at capture phase. Does it interfere with anything that already calls `stopPropagation` (sidebar row click, badge reveal click, settings tab click)? Is there a leak path where the popover is removed but the document listener stays?

4. **`tmuxCapture` argv.** We pass `tmux capture-pane -p -J -t <session>:<window> -S -<N>` with no `-e`, expecting tmux to strip ANSI. Is that assumption correct in current tmux versions, and could a window with `:` in its name (sanitized via `tmuxWindowName` to `-`) ever break the target spec? `clampHoverLines` clamps to `[1, 500]` — is the floor handling for negative input safe (returns 1 for ≤1, but what about NaN? Tested in `tmuxCapture.test.ts`)?

5. **Settings clamps.** `mergeSettings` rejects out-of-range numbers and falls back to defaults rather than clamping in place. Is that behavior consistent with the existing `recentResolvedLimit` pattern? Could the in-place editor (text input in settings tab) write a bad value that gets corrected only on next reload?

6. **Backward-compat with stored settings.** Existing users have no `agentHoverPreview` / `agentHoverLines` / `agentHoverDelayMs` in their saved JSON. `mergeSettings` should populate defaults — verified for fresh defaults, but is there any path where a partial `view: {}` override drops the new keys?

7. **`findAgentTmuxLocation` extraction.** I refactored `revealAgentSession.ts` to call the new helper, then do its own `select-window` + `activateTerminalApp`. Is the behavior identical to the prior monolithic implementation? Specifically: the prior `tmuxSelectWindow` returned `false` if `select-window` failed; now it returns `null` only if the window can't be found in any session, and a `select-window` failure is swallowed but still emits the "no live session" notice. Is that drift acceptable or should I preserve the old "select failed → try next session" behavior? (I think no — `select-window` against a confirmed-existing target is unlikely to fail for any reason that retrying would help with — but worth confirming.)

PR #183, opened 2026-04-25
OP-168: zellij-as-target review

earchibald:
## Summary

Research deliverable for OP-168 — review of [zellij](https://github.com/zellij-org/zellij) as a candidate replacement for the tmux + iTerm-orchestrator stack that hosts op-obsidian agents today.

**Verdict: HOLD.** Revisit when zellij closes the multi-client mirror gap ([#4253](https://github.com/zellij-org/zellij/issues/4253)) **or** the team consciously accepts losing the per-pane viewer pattern + iTerm-CC native tabs.

Key findings (full evidence in `docs/research/OP-168-zellij-as-target.md`):

- **Probe-verified.** Headless orchestration works today against zellij 0.44.1: `zellij attach -b` for detached create, `action new-tab -l <kdl>` for layout injection, `--pane-id` for `dump-screen`/`focus-pane-id` (focus-independent, no client required), `--tab-id` for cross-tab spawn, `list-sessions -s` as a clean has-session analogue.
- **Two structural gaps.** (1) Zellij is mirror-only when multiple clients attach to one session — the `view-${issueId}` grouped-tmux trick at `orchestrator.ts:263–290` doesn't translate. (2) No tmux -CC analogue → iTerm native-tabs-from-tmux integration (the non-orchestrator launch path) goes away.
- **Timing yellow flag.** The 0.44 CLI surface we'd depend on is ~3 weeks old. Zellij is still pre-1.0.
- **Migration size.** Role (c) — both replacements — sheds ~2.7k LoC of hand-written source + ~19k LoC of generated protobuf types. Significant, but not favorable enough today given the gaps and timing.
- **OP-154 is unaffected.** It should proceed independently and continues to improve the current stack.

## Test plan

- [ ] No code changes; doc-only PR. CI is green / N/A.
- [ ] Reviewer reads `docs/research/OP-168-zellij-as-target.md` end-to-end and sanity-checks the verdict against §3 probe transcripts and the dimension-by-dimension table.
- [ ] Reviewer challenges the LoC estimates by spot-checking the wc on the listed files.

### DIFF
--- a/docs/research/OP-168-zellij-as-target.md
+++ b/docs/research/OP-168-zellij-as-target.md
@@ -1,0 +1,231 @@
+---
+project: obsidian-projects
+type: doc
+doc_type: research
+issue: "[[OP-168 Review zellij as target]]"
+created: 2026-04-25
+status: draft
+tags:
+  - project/obsidian-projects
+  - doc
+---
+
+# OP-168 — Zellij as a target for the tmux + iTerm orchestrator
+
+Review of [zellij](https://github.com/zellij-org/zellij) as a candidate replacement for the tmux + iTerm-orchestrator stack that today hosts op-obsidian agents (`plugins/op-obsidian/src/terminalLaunch.ts`, `orchestrator.ts`, `iterm/*`). Evidence is mixed `[probe]` (verified locally with zellij 0.44.1 / `brew install zellij` on macOS 25.3) and `[doc]` (zellij.dev docs, GitHub source/issues).
+
+---
+
+## TL;DR — verdict
+
+**HOLD.** Revisit when **either** (a) zellij ships independent-per-client tab focus on a shared session (issue [#4253](https://github.com/zellij-org/zellij/issues/4253), open) **or** (b) the team consciously accepts losing the per-pane viewer pattern *and* iTerm's native-tabs-from-tmux integration. The probe demonstrates that the headless orchestration primitives we'd need (detached create, by-pane-id targeting, layout injection) all work today on 0.44.1, so the technical direction is sound. The blockers are not capability gaps in the small — they are loss of two specific architectural properties that today's stack relies on, plus the fact that the relevant CLI surface (`attach -b`, `action new-tab -l`, cross-session `--pane-id`) all landed in the just-shipped 0.44 release (April 2026). Pre-1.0 software with a 3-week-old automation surface is not the right base for a one-shot multi-week migration.
+
+OP-154 (migrate `terminalLaunch.ts` off AppleScript onto the iTerm WS wrapper) is **not affected by this verdict** and should proceed independently — it improves the current stack regardless of any future zellij decision.
+
+---
+
+## What today's stack actually does
+
+The orchestrator surface is bigger than the issue body's "~1.5k LoC" estimate:
+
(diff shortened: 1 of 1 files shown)


earchibald:
@copilot — adversarial review requested. Concrete pressure-tests:

1. **Verdict defensibility.** Is HOLD the right call, or is REJECT more honest given (a) zellij is pre-1.0, (b) the relevant CLI surface (`attach -b`, `action new-tab -l`, cross-session `--pane-id`) all landed in the just-shipped 0.44 release, and (c) the multi-client mirror gap is a structural feature absence, not a bug? Push back if you think the doc is hedging.

2. **LoC delta is overstated or understated.** The doc claims role (c) sheds ~2.7k LoC of hand-written source plus ~19k LoC of generated protobuf types. Spot-check the wc:
   - `plugins/op-obsidian/src/iterm/*.ts` (hand-written, excluding `proto/`): claimed 1057 LoC.
   - `plugins/op-obsidian/src/iterm/proto/api.generated.d.ts`: claimed 19178 LoC.
   - `terminalLaunch.ts` (321), `orchestrator.ts` (394), `agentSessionCleanup.ts` (135), `revealAgentSession.ts` (84), `staleAgentBadges.ts` (74), `tmuxDetect.ts` (25), `layout/layouts.ts` (144).
   What's missing? The doc explicitly excludes `iterm/*.test.ts` (523 LoC) from the "source" total — fair? Are there orchestrator tests not in the table? Is the `~400 LoC added` for a zellij driver realistic, or hand-wavey?

3. **Grouped-session attach analysis is correct?** The doc claims zellij's mirror-only multi-client model breaks the `view-${issueId}` grouped-tmux pattern at `orchestrator.ts:263–290`. The proposed workaround is "one zellij session per agent" instead of one shared session with N tabs. Does that workaround actually work, or does it lose something else (e.g., the ability to `zellij list-sessions` and see all agents at once for `staleAgentBadges` analogue)? What about the cleanup story — is N sessions × 1 agent each really cheaper to manage than 1 session × N tabs?

4. **§3 probe transcripts support the verdict.** The probe verified dimensions 3, 4, 6 (layout, external scripting, observability) and read-only confirmed 5, 7, 9, 10 from docs. Is the probe coverage sufficient, or did the doc skip a dimension worth verifying locally? Specifically — should the probe have actually launched two clients to *prove* the mirror-only behavior rather than relying on issue [#4253](https://github.com/zellij-org/zellij/issues/4253)?

5. **OP-154 / OP-155 interaction.** The doc says OP-154 is unaffected and should proceed. Is that right, or does landing OP-154 first make the eventual zellij decision harder (more invested in iTerm WS) and tip the cost calculus toward "reject" rather than "hold"?

6. **What's missed entirely?** A hostile reviewer's job is to surface dimensions the author didn't think about. Specifically: zellij plugin/WASM extension surface (would we want to write an Obsidian-aware status plugin?), session-resurrection semantics for crashed agents, signal handling differences (does Ctrl-C in a zellij pane propagate the same way as in a tmux pane to a Claude Code agent?), terminfo/`$TERM` value differences that might break TUI agents.

Goal: surface anything that should change the verdict from HOLD to either REJECT or RECOMMEND. If the verdict survives this review, ship it.

PR #184, opened 2026-04-25
OP-163: sidebar right-click context menu (§12 of OP-149)

earchibald:
## Summary

Right-click on any sidebar row opens an Obsidian `Menu` with state-conditional items: `Resolve…`, `Resolve as wontfix…`, `Open GitHub issue`. Each item dispatches an existing flow through the existing confirmation modal — no destructive shortcuts.

`Reopen` and `Detach agent` are wired in the menu builder but gated on `hooks.reopenIssue` / `hooks.detachAgent`, which `main.ts` leaves `undefined` for now. A future `op-reopen-issue` PR and §5's `op: detach agent` command light those items up automatically.

Item visibility is decided by a pure helper `buildSidebarMenuItems(entry, hooks)` so the conditional matrix is unit-testable without an Obsidian `Menu` DOM.

Bumps `op-obsidian` to `0.54.0` (additive feature, no breaking change).

## Test plan

- [x] `vitest run` — 594 plugin tests pass, including 9 new `buildSidebarMenuItems` cases.
- [x] Build via `node scripts/bump-version.mjs minor` (lockstep manifest/package/plugin.json + `npm run build`).
- [x] Sync into Agent-Vault and reload via `obsidian plugin:reload id=op-obsidian`.
- [x] Real right-click on a sidebar row in the active vault renders the menu (user-confirmed).
- [ ] Reviewer: verify Resolve / Resolve as wontfix items dispatch through `runResolve`'s confirmation modal (no double modal regression).
- [ ] Reviewer: verify Open GitHub issue opens the URL in the system browser.

## Spec

`docs/specs/OP-149-feel-great-ux-design.md` §12 (RECOMMEND right-click menus; REJECT drag-to-resolve and drag-to-reorder).

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.55.0",
+  "version": "0.56.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.55.0",
+  "version": "0.56.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/main.ts
+++ b/plugins/op-obsidian/src/main.ts
@@ -336,7 +336,11 @@ export default class OpPlugin extends Plugin {
             tmuxSessions: () => tmuxSessionsForCleanup(this.settings.orchestratorState),
             recordRecency: (id) => this.recordRecency(id),
             executeResumeLast: () => void this.runResumeLastCommand(),
-            resolveIssue: (entry) => this.runResolveCommand({ path: entry.path }),
+            resolveIssue: (entry, status) =>
+              this.runResolveCommand({ path: entry.path, status }),
+            openGithubIssue: (entry) => {
+              if (entry.githubIssue) window.open(entry.githubIssue, "_blank");
+            },
           },
           async (entry) => {
             const loc = await findAgentTmuxLocation(this.settings, entry.id);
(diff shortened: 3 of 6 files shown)


earchibald:
@copilot adversarial review please. Pressure points:

1. **Menu visibility matrix.** `buildSidebarMenuItems` decides visibility from `(entry.status, entry.resolvedFolder, entry.agent, entry.githubIssue)` × hook presence. Are there state combinations I'm missing? E.g. an entry with `status: "in-progress"` but `resolvedFolder: true` (file moved manually back from RESOLVED ISSUES/, frontmatter stale) — what should the menu show?

2. **Hook signature change.** `OpSidebarHooks.resolveIssue` went from `(entry) => …` to `(entry, status?: ResolveStatus) => …`. The keyboard `r` path in `resolveSelected()` still calls `hook(entry)` with no status — does the optional default to `"resolved"` propagate correctly through `runResolve` (status defaults to `"resolved"` at line 34 of `resolve.ts`)?

3. **Confirmation modal flow.** Resolve via right-click goes: contextmenu → `hooks.resolveIssue(entry, "resolved")` → `runResolveCommand({path, status})` → `runResolve` → its own confirmation modal. The keyboard `r` path adds a *second* `OpResolveConfirmModal` *before* `runResolve`'s modal. Is the contextmenu path's single-modal flow intentional and consistent with `op-resolve` palette command? (I think yes — keyboard `r` adds the extra modal because it's auto-repeat-prone.)

4. **Synthetic event filtering.** `Menu.showAtMouseEvent` is a no-op for synthetic dispatches (Obsidian filters by `evt.isTrusted` or similar). My smoke test confirms the handler fires on synthetic events but the menu only renders for real user clicks. Verified manually in the active vault. Does this affect any test environment?

5. **`stopPropagation` on contextmenu.** I call `ev.preventDefault(); ev.stopPropagation()` in the row's contextmenu handler. Does this clobber any ancestor handler (e.g. Obsidian's own workspace contextmenu, or a parent listener that contributes "Reveal in file explorer" / "Open in new pane" style items)? Should I leave the event bubbling so users get the union of menus?

6. **Open GitHub item parity.** The row's `GH #N` link in the meta strip already opens the issue on click via `window.open(url, "_blank")`. The new contextmenu item duplicates this. Worth keeping for discoverability or noise?

7. **Reopen / Detach agent placeholders.** Items are gated on `hooks.reopenIssue` / `hooks.detachAgent` which main.ts leaves undefined. Confirm that an undefined hook silently omits the item (no `Cannot read properties of undefined` at click time) — the type system should enforce this but worth a sanity check.

8. **Test coverage gaps.** `buildSidebarMenuItems` is well-covered; the contextmenu *wiring* in `render()` (line 295–301) is not directly tested. Is the existing `render()` test surface enough, or should I add a test that asserts a `contextmenu` listener was registered on the row?

PR #187, opened 2026-04-25
OP-156: persist agent badge across resolve when tmux session is alive

earchibald:
## Summary

- §5 of OP-149 — `op-resolve` keeps `agent:` set when the issue's tmux window is still live, surfacing an actionable Notice (`[Open agent]` / `[Detach]`) instead of clearing the badge out from under a still-attached session.
- New `op: detach agent` palette command + URI handler (`obsidian://op-detach-agent?id=<ID>`) + CLI handler. Idempotent kill-window + clear-`agent:` for crash zombies and explicit cleanup. Wires through the row context menu's `[Detach]` item that landed (as a hook stub) in OP-163.
- In flight sidebar tab keeps showing resolved-but-live rows (muted/strikethrough title + `resolved` chip) gated by tmux liveness — so the user can find their way back to a still-attached session even after `op-resolve`. Unknown/missing tmux info never demotes a row (matches the no-false-positive rule of `isAgentBadgeStale`).

## Implementation notes

- TOCTOU-safe: a single `tmux list-windows -t <session> -F '#W'` per session via the existing `probeLiveTmuxWindows` helper. No separate `has-session` probe.
- SessionEnd-wins race resolution: `runResolve` only retains `agent:` when the probe says alive at read time. If SessionEnd fires later, its URI handler runs `clearAgentOnIssue` on the moved file. Final state is correct regardless of order; no locks.
- tmux unreachable (binary missing / timeout) → keep `agent:` set + warn the user. Resolve is never blocked on tmux.
- New `agentKept` / `agentProbeOk` fields on `ResolveResult` and the `op-resolve` JSON payload so callers (and the resolve-time Notice) can distinguish "still attached" vs "tmux unreachable".

## Test plan

- [x] Unit: `resolve.test.ts` — alive→keep, dead→clear, probe-fail→keep+warn, probe-throws→keep+warn, no-agent→no probe.
- [x] Unit: `detachAgent.test.ts` — kill+clear, found=false, idempotent zero-kill, error surfacing.
- [x] Unit: `sidebarView.test.ts` — `pickIssuesForTab` for In flight resolved-with-agent rows (live / dead / unknown probe states).
- [x] Full vitest suite: 50 files, 634 tests passing.
- [x] Smoke: plugin reloads at 0.57.0; `op-detach-agent` registered; `withAgentProbe` correctly reports alive=true for the active OP-156 tmux window and alive=false for a synthetic id; `handleOpDetachAgentUri` returns `found=false` for an unknown id without touching state.

## Files changed

- `plugins/op-obsidian/src/resolve.ts` — accept optional `probeAgentLive`; conditionally retain `agent:` / `agent_session:`; expose decision on result.
- `plugins/op-obsidian/src/main.ts` — `withAgentProbe` wrapper, resolve-time §5 Notice, `op-detach-agent` command + URI + CLI, sidebar context-menu wiring.
- `plugins/op-obsidian/src/detachAgent.ts` — new module; `cleanupAgentSessions` + `clearAgentOnIssue`.
- `plugins/op-obsidian/src/sidebarView.ts` — extract `pickIssuesForTab` (pure), gate resolved+agent rows by liveness in In flight, render strikethrough + `resolved` chip.
- `plugins/op-obsidian/styles.css` — `.op-sidebar__item--resolved-live`, `.op-sidebar__resolved-chip`.
- `plugins/op-obsidian/manifest.json`, `plugins/op-obsidian/package.json`, `plugins/op/.claude-plugin/plugin.json` — bump 0.56.0 → 0.57.0.


earchibald:
@copilot Adversarial review please. Concrete pressure-test prompts:

1. **TOCTOU race between read-time probe and `processFrontMatter` write.** `runResolve` reads tmux liveness, then writes frontmatter, then renames the file. If the agent's tmux window dies mid-write (between probe return and `processFrontMatter`), we keep `agent:` set on the moved file. Is that acceptable? The spec says "SessionEnd-wins" — but SessionEnd's URI handler resolves the issue id by `store.issues().find(e => e.id === id)`. After the rename, the file is now in `RESOLVED ISSUES/` — does the store re-emit and reflect the post-rename path before the SessionEnd hook fires, or could SessionEnd race with the rename and miss the moved file? Walk through the worst-case timing.

2. **`cleanupAgentSessions` mutates settings.orchestratorState in place.** `detachAgent` calls `await args.saveSettings()` only when something changed. If two `detachAgent` calls run concurrently for different ids, both mutate the same registry object — last write wins. Is the existing `cleanupAgentStateFor` callsite single-threaded? Are we introducing a new race here?

3. **Sidebar in-flight filter and the probe-update lifecycle.** `pickIssuesForTab("in-flight", { liveTmuxWindows: undefined })` returns resolved+agent rows unchanged. The first probe lands ~immediately on `onOpen`. But if a user resolves an issue and the SessionEnd hook fires immediately after (clean exit), the order is: (a) frontmatter write keeps `agent:` because liveness was true at probe time, (b) rename, (c) SessionEnd URI clears `agent:` on the moved file. Between (a) and (c), the In flight tab briefly shows a strikethrough resolved-live row. Is that the desired UX, or should we wait?

4. **Notice action wiring.** The `[Detach]` action on the resolve-time Notice calls `runDetachAgentCommand(issueId)` directly. If the user clicks `[Detach]` _twice_ (rapid double-click), do we get two `cleanupAgentSessions` calls racing each other? `Notice` dismisses on the first click, so the second click won't fire — but what if Obsidian's actionable-notice has a brief overlap window?

5. **Resolve path with `closeGithubIssueOnResolve`.** I wrap `args.probeAgentLive` via `withAgentProbe(withGhCloseHook(args))`. The order matters: gh close runs in `onAfterMove` (after rename), but `probeAgentLive` runs before frontmatter write. Is the wrapper composition order correct? Could a caller-supplied `probeAgentLive` shadow `withGhCloseHook`'s `onAfterMove`?

6. **`withAgentProbe` short-circuits when `args.probeAgentLive` is already set.** The CLI handler `handleOpResolveCli` doesn't set it explicitly — so it picks up the live tmux probe. Is that what we want for an agent-driven CLI flow, or should CLI callers always skip the probe (since agents typically resolve their own issue and we _want_ the badge cleared)?

7. **`sidebarView.ts` `isResolved` helper duplication.** The helper appears at the top-level (used by `pickIssuesForTab`) and as part of the row-render logic (`resolvedLive`). Are the two semantics identical, or could they drift — e.g., does `e.resolvedFolder` ever differ from `e.status === "resolved" || e.status === "wontfix"` after a partial rename?

8. **Schema impact on resolved frontmatter.** Pre-PR, every resolved issue note had `agent:` cleared. Post-PR, some resolved notes will retain `agent:`. Does any downstream consumer (Bases, sidebar resolved tab, the all-projects aggregate, the `commits:` post-resolve back-fill, the GH issue mirror) assume the absence of `agent:` on a resolved note? List anything that breaks.

earchibald:
Thanks @copilot. Walked through all eight findings:

- **Q1 (TOCTOU + cache race)** — accepting the analysis: SessionEnd can land in the metadataCache→IssueStore refresh gap, leaving `agent:` permanently pinned on the moved file. Closed in **165375a** by polling the store lookup in `handleOpAgentEndedUri` (5×50ms = 250ms cap). The metadataCache `changed` event for the moved file typically lands in <50ms; the cap is well over the worst-case observed window without blocking the URI handler perceptibly. SessionEnd-wins is now provably correct, not aspirational.
- **Q2, Q3, Q4, Q5, Q7** — accepted as-is. The single-threaded JS guarantee, the deliberate brief-strikethrough UX, the `Notice.hide()` semantics, the wrapper composition's independent-fields property, and the helper duplication's identical semantics all check out.
- **Q6** — comment fix in **63f71df** matches the code. Behavior was always correct; the docstring drifted.
- **Q8** — accepted. The downstream-consumer audit is clean. Comment update in 63f71df documents the new `selectStaleAgentBadges` contract.

Bumped 0.57.0 → 0.57.1 (patch — defensive race fix, no public surface change). Smoke-tested locally; full suite still green.

PR #189, opened 2026-04-26
OP-151 + OP-161 + OP-162: note chip + status strip + first-run README (§2/§10/§11)

earchibald:
## Summary

Bundled per §16 of [[OP-149]] — all three issues share CM6 widget infrastructure with `AbortController`-based teardown. Closes OP-151, OP-161, OP-162.

- **§2 ([[OP-151]])**: primary-action chip above the H1 of every issue note. Label flips with state — \`▶ Start agent\` → \`▶ Attach session\` → \`↻ Re-attach (start fresh)\` → \`↺ Reopen\`. One \`⋯\` overflow menu per state, dispatched via \`app.commands.executeCommandById\`.
- **§11 ([[OP-162]])**: inline status strip below the chip — last commit (sha7 + subject), PR (#n + state), GH issue (#n + state). \`gh\` state lazy-fetched with a 60s in-memory cache; failures render the link without state and never block initial paint. Settings toggle \`view.disableInlineGithubStatus\` suppresses PR / GH segments for users without \`gh\`.
- **§10 ([[OP-161]])**: first-run README scaffolded into \`Projects/_op-readme.md\` on the very first plugin load. Contains the hotkey-preset table with an inline \`Apply preset\` chip and a \`Start tour\` chip that scaffolds a demo project (\`op-demo\`, prefix \`OPD\`) with three pre-seeded issues. Demo teardown chip lives on the demo project's \`STATUS.md\`. \`firstRunCompleted\` flag flips after the README is written so a manual delete is durable.

### Architecture notes

- **Chip mount.** A CM6 \`ViewPlugin\` mounts a chip-host \`<div>\` as a sibling of \`cm-editor\` (outside CM6's managed DOM tree). Obsidian's bundled CM6 rejects \`Decoration.widget({block: true})\` from both a \`ViewPlugin\` and a \`StateField\` provider — its facet marks both as plugin sources. Mounting outside the editor avoids the rule entirely and matches the pattern Obsidian uses for embedded markdown cards. Reading mode goes through a \`registerMarkdownPostProcessor\` that finds the first H1 and prepends a sibling chip wrapper.
- **Lifecycle discipline.** Every event listener attached in the chip's \`render\` path is bound to an \`AbortController\` stored on the widget / \`MarkdownRenderChild\`. \`destroy()\` / \`onunload()\` calls \`controller.abort()\` so listeners disappear deterministically across mode toggles. This is the single most likely bug class for this kind of widget.
- **Liveness.** \`isAgentLive(id, agent)\` is answered synchronously from a plugin-level \`liveTmuxWindowsCache\`, populated by the OP-150 startup probe and refreshed every 20s via a registered interval. When the live set changes, we dispatch a \`refreshChipEffect\` to every CM6 view so the chip flips promptly.
- **GH state cache.** Single \`Map<url, { state, expiresAt }>\` shared across renderers, 60s TTL, single-flight per URL (we plant a short-TTL placeholder when a fetch starts so 6 renders for the same URL collapse to 1 \`gh\` call).
- **\`op-action\` codeblock.** New fenced-code language used by the README chips. Body is a YAML-shaped key/value: \`action: <command-id>\` + \`label: <text>\`. The post-processor renders it as a chip wired to \`app.commands.executeCommandById\`.
- **New commands**: \`op-reopen-issue\`, \`op-clear-agent\`, \`op-set-priority\`, \`op-apply-preset\`, \`op-start-tour\`, \`op-remove-demo\`. The first three power the chip's overflow menu; the latter three back the README chips.
- **New settings**: top-level \`firstRunCompleted: boolean\` (default false); \`view.disableInlineGithubStatus: boolean\` (default false).
- **Liveness probe re-uses OP-150's \`probeLiveTmuxWindows\`** rather than a new \`tmuxLiveness.ts\` module — that file never landed in OP-150, but the existing probe + \`tmuxWindowName(id)\` mapping is sufficient.

### LOC + tests

- New: \`noteChipState.ts\` (193), \`noteStatusStrip.ts\` (180), \`noteDecorations.ts\` (480), \`noteChipMenu.ts\` (60), \`firstRunReadme.ts\` (260), plus 49 new unit tests across three test files.
- Modified: \`main.ts\` (+200, registers extension + post-processor + 6 commands + first-run hook + 20s liveness re-probe), \`settingsPure.ts\` (+10, two new fields + merge logic), \`styles.css\` (+135, themed via Obsidian CSS variables).
- 667 tests total, 51 test files, all passing.
- TypeScript: no new errors (33 pre-existing baseline → 33).

### Adversarial review prompts (for Copilot)

@copilot — please pressure-test, prioritising in this order:

1. **Listener teardown.** Toggle Live Preview ↔ Reading mode rapidly five times on the same issue note. Are there stale listeners on the original DOM? Is \`controller.signal.aborted === true\` after \`destroy()\`?
2. **Frontmatter-change feedback loop.** A chip click runs a command that mutates frontmatter, which fires \`metadataCache.changed\`, which calls \`dispatchChipRefresh\`. Verify no infinite loop — the metadata cache event is debounced; chip render must be idempotent.
3. **Two-renderer drift.** Live Preview chip and Reading mode chip must look and behave identically. Do they? Check: \`primaryLabel\`, \`primaryCommand\`, \`menu\` items, \`disableInlineGithubStatus\` honoured on both paths.
4. **GH state cache thundering herd.** Open a vault with five resolved issues that all link to PRs in the same repo. Is the \`gh pr view\` command de-duplicated, or are we firing five concurrent shells per render?
5. **First-run race.** Boot Obsidian with a vault that has \`firstRunCompleted: false\`. Does the README scaffold exactly once, even if the plugin reloads mid-startup?
6. **\`op-action\` codeblock idempotency.** Re-render the same codeblock multiple times (typical Obsidian post-processor behaviour). Do we stack duplicate chips inside the fence, or does \`el.empty()\` keep it clean?
7. **Schema migration.** A user with \`data.json\` from v0.57.x — does \`mergeSettings\` populate \`firstRunCompleted\` and \`view.disableInlineGithubStatus\` with sane defaults?

## Test plan

- [ ] Open an open issue note in Live Preview → see \`▶ Start agent\` above the H1.
- [ ] Click chip → agent launches; chip flips to \`▶ Attach session\` after the next 20s probe.
- [ ] Switch to Reading mode → identical chip via the post-processor.
- [ ] Open a resolved issue → chip reads \`↺ Reopen\`; overflow shows the GH issue link.
- [ ] Open a TASKS note → no chip rendered (only \`type: issue\` notes).
- [ ] Squeeze a leaf below 320px → chip hides; widen → chip reappears.
- [ ] Long title → chip ellipses; H1 stays readable.
- [ ] First-run README appears on a fresh vault; \`Apply preset\` chip applies the preset; \`Start tour\` chip scaffolds \`op-demo\`.
- [ ] Delete the README → \`firstRunCompleted: true\` keeps it gone across reloads.

## Smoke test caveat

This vault has another community plugin (\`jira-bases\`) that throws \`Block decorations may not be specified via plugins\` during \`setViewData\` for any note containing certain frontmatter keys. The error is pre-existing and unrelated to this PR — it predates the chip extension and persists with op-obsidian fully unloaded. The chip mount code sidesteps decoration-system entirely (mounted as a sibling of \`cm-editor\`) so it cannot trigger the same path. Reading-mode mirror, README scaffold, codeblock chip, and command registration all verified live.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.58.0",
+  "version": "0.58.1",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.57.1",
+  "version": "0.58.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.57.1",
+  "version": "0.58.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.58.0",
+  "version": "0.58.1",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/firstRunReadme.test.ts
+++ b/plugins/op-obsidian/src/firstRunReadme.test.ts
@@ -1,0 +1,131 @@
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("obsidian", () => ({
+  TFile: class {},
+  normalizePath: (p: string) => p,
+}));
+
+import {
+  buildDemoIssueNote,
+  buildReadmeBody,
+  DEMO_ISSUES,
+  parseOpActionBlock,
+  scaffoldFirstRunReadme,
+  shouldScaffoldReadme,
+  type ReadmeWriter,
+} from "./firstRunReadme";
+
+describe("buildReadmeBody", () => {
+  it("includes the Apply preset and Start tour codeblocks", () => {
+    const body = buildReadmeBody();
+    expect(body).toContain("```op-action");
+    expect(body).toMatch(/action:\s*op-apply-preset/);
+    expect(body).toMatch(/label:\s*Apply preset/);
+    expect(body).toMatch(/action:\s*op-start-tour/);
+    expect(body).toMatch(/label:\s*Start tour/);
+  });
+
+  it("starts with an H1", () => {
+    expect(buildReadmeBody().startsWith("# ")).toBe(true);
+  });
+});
+
+describe("DEMO_ISSUES", () => {
+  it("has exactly three pre-seeded issues", () => {
+    expect(DEMO_ISSUES.length).toBe(3);
+    expect(DEMO_ISSUES.map((s) => s.number)).toEqual([1, 2, 3]);
+  });
+
+  it("renders an issue note with valid frontmatter", () => {
+    const note = buildDemoIssueNote(DEMO_ISSUES[0]);
+    expect(note).toContain("id: OPD-1");
+    expect(note).toContain("type: issue");
+    expect(note).toContain("status: open");
+    expect(note).toContain("project: op-demo");
+  });
+});
+
+describe("shouldScaffoldReadme", () => {
+  it("scaffolds when flag is false and file is absent", () => {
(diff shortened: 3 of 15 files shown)


earchibald:
@copilot — adversarial review please. Pressure-test the PR body's prompts in priority order; #1 (listener teardown) and #2 (frontmatter feedback loop) are the biggest blast-radius risks. Highlights to focus on:

- The chip mounts OUTSIDE CM6's managed DOM (sibling of `cm-editor`). Verify this doesn't break Obsidian's own scroll/sticky-header behaviour, or duplicate when leaves split.
- The 20s liveness re-probe runs unconditionally regardless of whether any issue note is open. Worth gating?
- `dispatchChipRefresh` iterates ALL leaves and dispatches a transaction on each — confirm this scales and doesn't fight CM6's autocomplete or selection state.
- The `op-action` codeblock processor uses `el.empty()` for idempotency — does that interact badly with Obsidian's own inline embedded views?
- `gh pr view --json state` is shelled per URL with an 8s timeout; on slow networks the chip stays in placeholder state forever after timeout. Should we negative-cache failures?
- Schema migration: a user upgrading from 0.57.x lacks `firstRunCompleted` and `view.disableInlineGithubStatus`. `mergeSettings` falls back to defaults — confirm no false-trigger of the first-run README on existing users.

Pre-existing vault issue noted in PR body (`jira-bases` blocks Live Preview for some issue notes) is NOT my code — verified pre-PR.

PR #191, opened 2026-04-26
OP-159: quick-capture — op: new from selection / clipboard (§8 of OP-149)

earchibald:
## Summary

Implements §8 of [docs/specs/OP-149-feel-great-ux-design.md](https://github.com/earchibald/obsidian-projects/blob/main/docs/specs/OP-149-feel-great-ux-design.md) — two thin quick-capture entry points around `op-new` that pre-fill the New Issue modal and skip the project picker when context resolves a project.

- **`op: new from selection`** (`editorCallback`) — uses `editor.getSelection()` as the issue's scope body, derives the title from the first line. No selection → title = active note's basename, scope = `Source: [[<note-title>]]`.
- **`op: new from clipboard`** — reads `navigator.clipboard.readText()`. Rejection (no focus / permission denied) → empty modal + one-shot Notice.
- Project resolution: (a) active note's `project:` frontmatter, (b) most-recent project from the recency log (prefix→slug lookup, advances past stale entries), (c) `ProjectSuggestModal` as last resort.
- Pre-fill mechanism: `NewIssueModal` now accepts optional `initial: { title, scopeRaw, scopeMode }`. `scopeMode: "body"` submits the textarea verbatim as `scopeBody` so paragraphs / blockquotes / blank lines round-trip intact instead of being shredded into one-bullet-per-non-empty-line by the legacy parser.
- Hotkey preset gains `⌘⌥S` (selection) and `⌘⌥V` (clipboard) sharing the `Mod+Alt+` family with `op-new`'s `⌘⌥N`. Preset count 10 → 12, all existing collision-skip semantics unchanged.

User-facing confirmation step is preserved per the op skill's "always pause before mutating vault or repo" rule — the goal is fewer keystrokes, not silent mutation.

Spec: `docs/specs/OP-149-feel-great-ux-design.md` §8.
Issue: `Projects/obsidian-projects/ISSUES/OP-159 …`

## Test plan

- [x] 706 unit tests pass — including 22 new tests on the pure `quickCapture` helpers (title derivation across markdown-noise variants, scope-body packing with/without fallback backlink, full (a)→(b)→(c) project resolution chain incl. stale recency entries and projects without prefixes).
- [x] Smoke-tested in Agent-Vault (Obsidian 1.8.x):
  - Clipboard variant: pre-filled title `"Test capture"` (markers stripped from `## Test capture`), scope textarea verbatim with `## heading + paragraph + bullets`. Project picker skipped.
  - Selection variant via injected fake editor: same prefill behavior.
  - No-selection fallback: title = active note's basename, scope = `Source: [[<note-title>]]`.
  - Console clean across all three runs.
- [x] Built artifact (`main.js`) deployed and reloaded; both commands present in `app.commands.commands`.

## @copilot — adversarial pressure-test prompts

Please pressure-test the diff against these specific concerns:

1. **Clipboard rejection paths.** `navigator.clipboard.readText()` can reject for at least three reasons: document not focused, user denied permission, no clipboard text at all. Trace the `runNewFromClipboardCommand` rejection path — is the user always told *why* the prefill didn't happen, or do silent-failure cases exist (e.g. resolved-with-empty-string vs reject)? Should the `text = ""` resolved-empty case also surface a Notice?
2. **Project-resolution staleness window.** `resolveCaptureProject` reads `app.metadataCache.getFileCache(file)?.frontmatter?.project`. What happens if metadataCache is mid-update for the active note (frontmatter recently edited)? Does the resolution silently use the stale slug, or is there a user-visible signal that the picker would have been more correct?
3. **Recency-log prefix collisions.** Two projects sharing a prefix (legacy projects pre-`prefix:`-field, manual edits, or test fixtures) → which one does `projects.find(p => p.prefix === prefix)` return? Is the answer deterministic across reloads given `listProjects()`'s sort order?
4. **`scopeBody` vs `scope` exclusivity in createIssue.** `issueTemplate.ts` says "if both are present, `scopeBody` wins." My modal sets `scope: undefined` when `scopeBody` is set, but verify the createIssue pipeline (esp. URI / CLI dispatch) handles the both-set case correctly — paste a paragraph into a prefilled modal, then add some bullets manually; what ships?
5. **Title-derivation pathological inputs.** What does `deriveTitle` produce on: (a) a 200-char single line of all `>>>>` blockquote markers, (b) a 200-char single line where the only non-marker char is at the end, (c) text starting with `​` zero-width space, (d) tab-indented bullets `\t- foo`? Are any of these a "looks empty but isn't" footgun?
6. **Modal pre-fill XSS / injection.** The pre-filled title goes through Obsidian's `setValue` on a text input — that's safe. But does `scopeBody` ever get rendered as HTML rather than text before `createIssue` writes it to a file? Specifically: any `innerHTML` / `setText` mistake in the textarea path?
7. **Hotkey preset bypass criteria.** The preset spec at OP-149 §3 calls out three bindings deliberately *swapped* away from Obsidian core conflicts. `⌘⌥S` and `⌘⌥V` — verify against current Obsidian 1.8 default hotkeys table that these are still free; if not, propose a non-conflicting alternative.
8. **No-selection fallback note-title escaping.** `Source: [[<basename>]]` — what if the active note's basename contains `]]`, `|`, or `#`? Does the wikilink still resolve, or do we need to wrap with `<...>` per Obsidian's escape rules?

Reply with concrete findings — line-numbered if possible. If you find a real bug, push the fix as a commit on this branch.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.58.1",
+  "version": "0.59.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.58.2",
+  "version": "0.59.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.58.2",
+  "version": "0.59.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.58.1",
+  "version": "0.59.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/hotkeyPreset.test.ts
+++ b/plugins/op-obsidian/src/hotkeyPreset.test.ts
@@ -12,8 +12,8 @@ import {
 describe("defaultPreset", () => {
   const preset = defaultPreset();
 
-  it("returns exactly 10 entries", () => {
-    expect(preset).toHaveLength(10);
+  it("returns exactly 12 entries", () => {
+    expect(preset).toHaveLength(12);
   });
 
   it("returns the expected commands in spec-defined order", () => {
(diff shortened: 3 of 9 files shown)


earchibald:
@copilot — please pressure-test this diff against the 8 concerns listed in the PR description. Concrete, line-numbered findings preferred. If you find a real bug, push the fix as a commit on this branch.

PR #192, opened 2026-04-26
OP-155: background-launch + cross-app focus return + iTerm prefs Notice (§4 Steps 1, 3, 4)

earchibald:
- Verify background-launch behavior really doesn't activate iTerm: trace the call graph from `backgroundLaunch=true` → `terminalLaunch.ts:launchInTerminal` → either `createTab(..., { activate: false })` or `createWindow(..., { activate: false })`. Are there *any* other code paths that could still bring iTerm forward (e.g. `setSessionName`, profile-property side-effects)?
- The orchestrator path: when growing into an existing iTerm window via `splitSession`, does the split-pane RPC have any inherent activate side-effect? My read is no — only `ActivateRequest` activates. Confirm.
- `iterm/cookie.ts` should be the *only* `osascript` call left in the bundle. Run `grep "osascript" plugins/op-obsidian/main.js` after building and confirm exactly one match. Is the cookie path reachable in this PR's flow without re-introducing scripted activation?
- The one-time Notice is gated on `iTermPrefsNoticeShown` AND `iTermDefaultsDomainPresent()`. Is there a race where `openAgent` returns before the (async) `maybeShowITermPrefsNotice` flips the bit, and a second concurrent launch double-fires the Notice? My take: not reachable because Obsidian commands are user-driven and `openAgent` itself is awaited per-invocation, but I want a second pair of eyes.
- Settings UI: I rendered the toggle inline under the existing Terminal section. Reasonable placement, or should it land closer to the orchestrator settings (since both gate iTerm focus behavior)?
- Cross-app `open -a "iTerm"` vs the legacy `iTerm2` (`tell application "iTerm2"`): I claim `open -a` matches the user-visible app name (LaunchServices), so "iTerm" is the right token. Verify against your understanding of LaunchServices resolution rules.
- Test coverage: the multi-message transport in `iterm/client.test.ts` asserts the activate→createTab pair becomes a single createTab when activate is false. Are there integration-level test gaps (e.g. terminalLaunch.ts wiring of `open -ga`) that should be unit-tested too?

## Summary
- **§4 Step 1 — Background-launch mode.** New `backgroundLaunch` setting + UI toggle. When on, `open -ga iTerm` cold-starts iTerm without activation; WS `CreateTab`/`CreateWindow` calls skip the `ActivateRequest`. Default off (current behavior).
- **§4 Step 3 — Cross-app focus return without `osascript`.** New `crossAppActivate.activateApp()` helper uses `open -a`. `revealAgentSession.ts` migrated off `osascript`; the only remaining osascript callsite in the bundle is `iterm/cookie.ts` (first-run auth, spec-sanctioned).
- **§4 Step 4 — One-time iTerm tmux-prefs Notice.** New `iTermPrefs.ts` surfaces an actionable Notice on the first iTerm launch listing the four pref recommendations from the spec, with an `[Open iTerm settings]` action button (`open -a iTerm`). Gated by persisted `iTermPrefsNoticeShown` + `iTermDefaultsDomainPresent()`.

Out of scope: §4 Step 2 (per-launch ⌥+click override) — OP-155 covers Steps 1, 3, 4 per its title.

Builds on top of OP-154 squash `6e34205` (AppleScript retirement). Net `osascript` callsite count: 1 (cookie.ts), unchanged from OP-154.

## Test plan
- [x] `npx vitest run` — 698 tests pass, including the 4 new client-activate-flag tests, 2 new driver pass-through tests, the iTermPrefs unit tests, and the crossAppActivate tests.
- [x] `npm run build` — production esbuild succeeds; bundle size 800K.
- [x] Smoke: `obsidian eval` confirms the new `backgroundLaunch` and `iTermPrefsNoticeShown` keys appear on `app.plugins.plugins["op-obsidian"].settings`, default to `false`, and round-trip through `saveSettings()`.
- [x] Smoke: Settings tab renders the new toggle (`Launch agent without stealing focus`) under the Terminal section.
- [x] Smoke: `dev:console` clean after plugin reload — no errors.
- [x] Smoke: `grep osascript plugins/op-obsidian/main.js` returns exactly 1 match (the cookie.ts site).
- [ ] Manual: flip `backgroundLaunch` on, launch a real agent, confirm Obsidian keeps focus and iTerm tab spawns in the background.
- [ ] Manual: clear `iTermPrefsNoticeShown` (`obsidian eval`), launch an iTerm agent, confirm the Notice fires with the `[Open iTerm settings]` action.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.58.2",
+  "version": "0.59.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.58.2",
+  "version": "0.59.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/crossAppActivate.test.ts
+++ b/plugins/op-obsidian/src/crossAppActivate.test.ts
@@ -1,0 +1,56 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+const execFileMock = vi.fn();
+
+vi.mock("child_process", () => ({
+  execFile: (file: string, args: string[], cb: (err: Error | null) => void) => {
+    execFileMock(file, args);
+    cb(null);
+  },
+}));
+
+import { activateApp } from "./crossAppActivate";
+
+describe("activateApp", () => {
+  beforeEach(() => {
+    execFileMock.mockClear();
+  });
+
+  it("invokes /usr/bin/open -a <name>", async () => {
+    await activateApp("Obsidian");
+    expect(execFileMock).toHaveBeenCalledWith("/usr/bin/open", ["-a", "Obsidian"]);
+  });
+
+  it("works for iTerm and Terminal targets", async () => {
+    await activateApp("iTerm");
+    await activateApp("Terminal");
+    expect(execFileMock).toHaveBeenNthCalledWith(1, "/usr/bin/open", ["-a", "iTerm"]);
+    expect(execFileMock).toHaveBeenNthCalledWith(2, "/usr/bin/open", ["-a", "Terminal"]);
+  });
+
+  it("never adds new osascript callsites", async () => {
+    await activateApp("Obsidian");
+    const calls = execFileMock.mock.calls;
+    for (const [bin] of calls) {
+      expect(bin).not.toContain("osascript");
+    }
+  });
+});
+
+describe("activateApp swallows failures", () => {
+  it("does not throw when execFile errors", async () => {
+    vi.resetModules();
+    vi.doMock("child_process", () => ({
+      execFile: (
+        _file: string,
+        _args: string[],
+        cb: (err: Error | null) => void,
+      ) => {
+        cb(new Error("boom"));
+      },
+    }));
+    const { activateApp: activateAppErroring } = await import("./crossAppActivate");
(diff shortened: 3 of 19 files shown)


earchibald:

1. **Background-launch really doesn't activate iTerm.** Trace `backgroundLaunch=true` through `terminalLaunch.ts:launchInTerminal` and `orchestrator.ts:orchestrateLaunch`. Are `createTab(..., { activate: false })` / `createWindow(..., { activate: false })` the only places that could bring iTerm forward? Anything else with an implicit activate side-effect (`setSessionName`, profile-property writes, the WS connection itself, the `iTerm` defaults probe)?
2. **Orchestrator growth path.** When growing into an existing iTerm window via `splitSession`, does iTerm's split-pane RPC have any inherent activate side-effect? My read: no — only `ActivateRequest` activates. The orchestrator's `selectSession` reuse path on re-launch *does* still activate, but that's intentional and matches today's behavior.
3. **Single remaining `osascript` callsite.** Run `grep "osascript" plugins/op-obsidian/main.js` after building — should be exactly 1 match, and only inside `iterm/cookie.ts`'s first-run auth path. Is there any reachable code in this PR that re-introduces a scripted activation?
4. **One-time Notice race.** `maybeShowITermPrefsNotice` is fire-and-forget (`void` from `openAgent`). Could two near-simultaneous `op-open-agent` calls double-fire the Notice before `iTermPrefsNoticeShown` flips? Realistically not — Obsidian commands are user-driven and `openAgent` itself is awaited per-invocation — but I want a second pair of eyes on the gate ordering.
5. **Settings sanitizer.** `mergeSettings` ignores any non-boolean value for `backgroundLaunch` / `iTermPrefsNoticeShown`. Is there a legacy data shape (older `settings.json`) where these should be preserved differently, or is "missing → default false" correct?
6. **`open -a "iTerm"` vs the legacy `iTerm2` AppleScript target.** I claim `open -a` matches the user-visible app name (LaunchServices), so "iTerm" is the right token; `osascript` historically wanted "iTerm2" because that's the bundle id. Confirm against your understanding of LaunchServices resolution rules.
7. **Test coverage gaps.** The new client-activate-flag tests assert the activate→createTab pair becomes a single createTab when `activate: false`. The `terminalLaunch.ts` wiring of `open -ga iTerm` isn't unit-tested — is that an acceptable gap (covered by the manual smoke test in the PR body) or should it be a unit test with a mocked `child_process.execFile`?

Out of scope: §4 Step 2 (per-launch ⌥+click override) — covered by a separate ticket.

Issue #14, opened 2026-04-22
once initial iTerm tmux window is started just add agents to tmux session without launching new tmux -CC


earchibald:
Tracked as op issue **OP-43**.

## Requirements

1. **Behavior**: Once the initial iTerm `tmux -CC` window is started and attached, subsequent agent launches must add windows to the existing shared tmux session without spawning a new `tmux -CC` attach (which creates a second iTerm window mirroring the session).

2. **Debug command**: Add an Obsidian command-palette entry `Obsidian Projects (op): open agent to debug agent launch` (or similar) that launches an agent window through the same code path but **without** passing an automatic-start prompt. This lets us manually exercise the launch flow and observe tmux session/window state without Claude immediately running.

Acceptance:
- Launching two agents in succession does not open a second iTerm window — the second shows up as a new tab/window inside the existing `tmux -CC` iTerm window.
- Debug command is visible in command palette, opens the agent window, and leaves the shell at an interactive prompt (no auto-run of Claude).

earchibald:
Resolved as OP-43 at 0.16.0. Shipped the debug-launch command (commits cfb229f, 79a2bbc). Full iTerm API layout orchestrator moved to OP-46 (#19).

Issue #12, opened 2026-04-22
update skill to reflect github issue linkage workflow (agent thought the github issue still remained open, plugin had closed it)


earchibald:
Tracked as op issue **OP-1**.

Issue #11, opened 2026-04-22
launched agent windows are missing statusline config that normal claude runs have


earchibald:
Tracked as op issue **OP-41**.

PR #193, opened 2026-04-26
OP-164: Settings tab — two-tier layout, search, targeted re-renders

earchibald:
## Summary

- Restructure `plugins/op-obsidian/src/settings.ts` (1028 LoC monolith) into ~10 `renderSection(containerEl)` methods with a Daily / Advanced two-tier layout, a fuzzy search box, JS-driven collapsibles, inline validation, and targeted per-section re-renders.
- Setting semantics unchanged — every toggle, dropdown, and validator behaves identically to 0.58.2. Pure UX restructure.
- Bumped 0.58.2 → 0.60.0 (minor: additive UX restructure of public surface, new exported `OpCollapsible` class).

## What changed

- **Daily group at top:** Default agent · Terminal app · Sidebar default tab · Onboarding README · Hotkey preset apply.
- **Advanced wrapper** with one collapsed-by-default subsection per H2: Injection · Working directories & project order (folded together) · iTerm layout orchestrator · Profile overlays · Agent worktree enforcement · Flow chaining · GitHub integration · Developer commands.
- **`OpCollapsible` primitive** uses JS-controlled `display: none|block`, NOT native `<details>` — closed `<details>` zeros out child `getBoundingClientRect()` and breaks the project-order drag logic. Collapsibles also refuse to collapse mid-drag (`canCollapse` predicate consults `dragInFlight`).
- **Per-section "What is this?" expandables** lift glossary terms next to each subsection header. The full glossary remains as a fallback `<details>` at the bottom of the tab.
- **Fuzzy search** uses `prepareFuzzySearch` (same primitive as the sidebar). `matchSettingRow` extracted to `settingsPure.ts` for testability. Filter hides non-matching rows, auto-expands matching collapsibles, hides collapsibles whose visible row count drops to zero.
- **Inline validation:** overlay textareas show green ✓ / yellow ⚠ / red ✗ banners on debounced `input` (150ms) + `blur`; tmux binary input shows a ✓/✗ `existsSync` status pill. Existing `userError` Notices preserved on save.
- **Project order folded into Working directories** — same project list, same concerns. Reset-to-alphabetical button bound inside `renderProjectOrder` with `rerenderSection("workingDirs")` instead of dangling outside the H2 with a full-tab redraw.
- **Targeted re-renders:** every existing `this.display()` call (Re-probe, Auto-detect tmux, Reset alphabetical, Add/Remove mapping, Bulk-edit paste) now repaints only its own section.
- **CLAUDE.md smoke-test recipe** updated for the new selector landscape (`.op-settings__group--daily`, `.op-settings__group--advanced`, `[data-op-section="<id>"]`); counting `.op-project-order__item` now requires a one-line collapsible expand first.

## Files

- `plugins/op-obsidian/src/settings.ts` — primary refactor (1029 → 1561 LoC, split into named functions).
- `plugins/op-obsidian/src/settingsPure.ts` — added `matchSettingRow` + `SettingMatcherFactory`.
- `plugins/op-obsidian/src/settings.test.ts` — 5 new vitest cases for `matchSettingRow`.
- `plugins/op-obsidian/styles.css` — new classes for search, collapsibles, help expandables, overlay/tmux validation banners.
- `CLAUDE.md` — smoke-test recipe + section-id list.
- `plugins/op-obsidian/manifest.json` · `package.json` · `plugins/op/.claude-plugin/plugin.json` — version bump to 0.60.0.

## Test plan

- [x] `npx vitest run` → 692/692 tests pass.
- [x] `npx tsc --noEmit` → no new errors (pre-existing AppLike + debug-duplicate noise unchanged).
- [x] `npm run build` → `main.js` rebuilt, fresher than `manifest.json`.
- [x] Vault-sync into Agent-Vault + `obsidian plugin:reload id=op-obsidian` → live at 0.60.0.
- [x] `app.setting.openTabById("op-obsidian")` → 5 Daily rows, 8 Advanced collapsibles, search input, glossary all rendered.
- [x] Expand `[data-op-section="workingDirs"]` → 10 draggable project rows, all `draggable=true`, first row's `getBoundingClientRect()` reports non-zero height/width (drag-in-collapsible bug avoided).
- [x] Search "tmux" → 8 hits including `tmux binary`, `Terminal app`, orchestrator, agent hover preview; clearing restores 42 visible rows.
- [x] Type invalid JSON in claude overlay textarea → red ✗ banner with parse error within 250ms; valid JSON → green ✓ "Valid overlay".
- [x] No console errors after reload.

## Spec

`docs/specs/OP-149-feel-great-ux-design.md §13` — drives the two-tier layout, search, inline validation, and targeted re-render asks.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,10 +43,19 @@ Always, in order:
 5. **Smoke-test Settings UIs.** When the change touches the op-obsidian Settings tab — Project order, working dirs, agent overlays, flow chaining, GitHub integration — `executeCommandById` won't reach it. Use `app.setting.openTabById` to render the tab synchronously and assert against the live DOM:
 
    ```bash
-   obsidian eval code='(()=>{ app.setting.open(); app.setting.openTabById("op-obsidian"); return document.querySelectorAll(".op-project-order__item").length; })()'
+   # Daily-group rows (Default agent / Terminal / Sidebar tab / Onboarding / Hotkey preset) — visible from open:
+   obsidian eval code='(()=>{ app.setting.open(); app.setting.openTabById("op-obsidian"); return document.querySelectorAll(".op-settings__group--daily .setting-item").length; })()'
+
+   # Advanced subsections (each is its own collapsible) — count the wrappers:
+   obsidian eval code='(()=>{ app.setting.open(); app.setting.openTabById("op-obsidian"); return document.querySelectorAll(".op-settings__group--advanced .op-collapsible").length; })()'
+
+   # Project-order drag rows: post-OP-164 they live INSIDE the "Working directories & project order" collapsible, which starts collapsed. Expand it first by clicking the header, then count:
+   obsidian eval code='(()=>{ app.setting.open(); app.setting.openTabById("op-obsidian"); document.querySelector("[data-op-section=\"workingDirs\"] .op-collapsible__header").click(); return document.querySelectorAll(".op-project-order__item").length; })()'
    ```
 
    Returns synchronously — no `await` needed. Use it to verify drag-row count, `draggable=true` flags, prefix badges, reset-button presence, etc. without touching the GUI. Worked example: after a change to the "Project order" section, the count above should equal the number of discovered projects (one `.op-project-order__item` per project). Swap the selector for the section you actually changed.
+
+   **OP-164 layout note.** Sections are tagged with `data-op-section="<id>"` on the collapsible wrapper. IDs are: `injection`, `workingDirs`, `orchestrator`, `profileOverlays`, `worktreeEnforcement`, `flowChaining`, `github`, `developer`. Each starts collapsed; click its `.op-collapsible__header` to expand before asserting against rows inside. The fuzzy search box at the top auto-expands matching collapsibles when its `value` is set and an `input` event is dispatched.
 
    Cross-reference: OPD-8 captures the generic obsidian-plugin-creator skill update for this recipe (also covers the async-modal-probe hang). The snippet here is the OP-side mirror so agents working op tasks aren't blocked on the OPD skill being updated first.
 
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.59.0",
+  "version": "0.60.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.59.0",
+  "version": "0.60.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/settings.test.ts
+++ b/plugins/op-obsidian/src/settings.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { DEFAULT_SETTINGS, mergeSettings } from "./settingsPure";
+import { DEFAULT_SETTINGS, mergeSettings, matchSettingRow } from "./settingsPure";
 import { emptyRegistry, mergeRegistry, type RegistryData } from "./layout/registry";
 
 describe("mergeSettings", () => {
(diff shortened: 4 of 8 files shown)


earchibald:
@copilot — adversarial review please. This is a pure UX restructure of the Settings tab; setting *semantics* must not change. Concrete pressure-test prompts:

1. **Drag-inside-collapsible regression.** `OpCollapsible` uses JS-controlled `display: none|block` instead of native `<details>`. Confirm the project-order drag list still works when (a) the user expands the "Working directories & project order" collapsible *then* drags, (b) the user starts a drag inside an already-open collapsible and the user clicks the header mid-drag — the `canCollapse` predicate refuses to collapse while `dragInFlight` is true; verify nothing else can flip `dragInFlight` back to `false` before `dragend` fires. What happens if the drag is cancelled (Escape) without a `dragend`? Does the flag leak?
2. **`getBoundingClientRect()` on hidden bodies.** Even though we don't read rects inside `display: none` bodies, do any of the `dragover`/`drop` listeners on `.op-project-order__item` ever execute against a row whose ancestor is `display: none`? If so, the rect is zero.
3. **Targeted re-render coverage.** Every `this.display()` call was replaced. Did any get missed? Cross-check `Re-probe`, `Auto-detect tmux`, `Reset alphabetical`, `Add mapping`, `Remove mapping`, `Bulk-edit JSON paste`, and `Detection`.
4. **`rerenderSection` cache invalidation.** `this.sectionEls` is populated in `mountSection`. After a rerender, the wrapper element identity is the same — but inner closures (e.g. `newSlug`/`newPath` captured in `Add mapping`) are torn down. Are there any outstanding references that survive across rerenders and write into stale closures?
5. **Search filter visibility.** `applyFilter` walks `.setting-item` rows and toggles `display`. Could it accidentally hide rows in the modal's hotkey-results dialog (which renders inside the tab via `new HotkeyPresetResultsModal(...)` opened from the Daily group)? Check whether `.setting-item` selectors leak across modal boundaries.
6. **Filter persistence across rerenders.** `applyFilter` runs after every `rerenderSection`. If a user is searching "tmux" and clicks "Auto-detect", the orchestrator section gets re-rendered — does the filter re-apply correctly? Are the `display: none` styles preserved on the new rows?
7. **Help-expandable z-index / focus.** The "What is this?" expandables sit inside `OpCollapsible__body`. Click handlers don't `stopPropagation` — does clicking the help expandable inside a collapsible accidentally also toggle the parent collapsible? Check the bubbling chain.
8. **Inline overlay validation race.** The debounced 150ms `input` validator and the synchronous `blur` handler can fire close together. If the user types invalid JSON and tabs away within 50ms, does the order of operations leave a stale banner visible?
9. **`existsSync` in the renderer.** The tmux binary path validator calls `existsSync` on every `input` event (debounced). On a slow network volume this could block the UI thread. Should this be queued differently?
10. **Search query DoS.** A fuzzy match against ~40 settings × `prepareFuzzySearch` is fine, but if the user pastes a 500-char query, does the matcher gracefully reject or thrash?
11. **Backwards compat / data.json.** No new settings keys, no removed keys, no schema changes — but does `mergeSettings` still produce identical output for an existing user's `data.json`? `OP_TEST` vault on a stale 0.58.x install should boot to identical settings state.
12. **Accessibility.** Native `<details>` was the prior glossary primitive; the new `OpCollapsible` is a div. Does it expose `role="button"` / `aria-expanded` for screen readers, and does focus + keyboard activation work (Enter/Space)? Right now I only wired click — verify.
13. **TS error budget.** `npx tsc --noEmit` count is unchanged from pre-PR baseline (pre-existing `AppLike` + `debug` duplicates). Confirm the diff doesn't *introduce* any new errors masked by the pre-existing ones.
14. **`SectionId` exhaustiveness.** The `dispatchRender` switch is exhaustive over `SectionId`. If a future PR adds a new section ID without updating `dispatchRender`, does the switch error at compile time, or silently fall through?

Specifically curious about #1, #4, #11, #12.

earchibald:
Pulled @copilot's fix commit (`9690e64`), reviewed it, and verified the four code changes:

- **#6 applyFilter body-visibility bug** — confirmed real bug. The CSS-only visibility model (`.op-collapsible__body { display: none }` + `.op-collapsible.is-open > .op-collapsible__body { display: block }`) is the right shape; `data-op-auto-expanded` cleanly distinguishes user-opened from filter-opened collapsibles for the clear-search restore path.
- **#12 accessibility** — `role="button"`, `tabindex="0"`, `aria-expanded`, Enter/Space keydown handlers, `:focus-visible` outlines all good.
- **#14 exhaustiveness** — `AdvancedSectionId = (typeof ADVANCED_SECTIONS)[number]["id"]` + `default: { const _exhaustive: never = id; return _exhaustive; }` lands the compile-time guarantee I asked for.
- **#1 dragInFlight safety net** — document-level `{capture: true, once: true}` listener is appropriately defensive without leaking listeners.

Live smoke-test on Agent-Vault confirms the bug fix: searching "tmux" auto-expands 3 collapsibles whose bodies are now actually visible (previously stayed hidden via inline `display: none` overriding `is-open`). Clearing the query restores all 8 to collapsed.

Caught one small a11y inconsistency in the fix and pushed `4cfba4c`: `applyFilter` was setting `aria-expanded` on the `.op-collapsible` wrapper, but OpCollapsible's constructor + `setOpen` both target the `.op-collapsible__header` (the actual `role="button"` element). Net effect was that screen readers saw the header still reporting `aria-expanded="false"` after auto-expansion. Moved the writes to the header so all three sites converge.

Verified post-fix:
- Search "tmux" → 3 auto-expanded, header `aria-expanded="true"` on all 3.
- Clear query → 0 auto-expanded, header `aria-expanded="false"` on all, no `is-open` class.

Ready to merge. All 717 tests pass. No new TS errors.

PR #194, opened 2026-04-26
OP-147: OP-Test test vault + seed-tag reset scripts

earchibald:
## Summary

- Pivot dev sync from Agent-Vault to a clean-room test vault at `~/Documents/OP-Test/OP-Test/` (Agent-Vault becomes a BRAT customer of op-obsidian, post-merge).
- Add four new scripts + a shared helper for OP-Test workflows:
  - `dev-sync.mjs` (replaces the manual cp+reload dance, with active-vault assertion + Agent-Vault guard + first-install fallback)
  - `build-seeds.mjs` (idempotent seed-ladder builder: `empty → scaffolded → mid-flow → github-linked → multi-project`; ensures the `op-test-fixture` private GH repo idempotently)
  - `reset-test-vault.mjs <seed>` (`git reset --hard seed/<name> && git clean -fd`, then plugin reload)
  - `agent-vault-detach.mjs` (one-shot, idempotent removal of dev artifacts from Agent-Vault, preserving `data.json`)
  - `lib/op-test.mjs` (single-source-of-truth for the OP-Test paths + the active-vault realpath assertion + obsidian/git CLI wrappers)
- Retarget CLAUDE.md: replace the `Sync into the active vault` step with `node scripts/dev-sync.mjs`, rewrite the OP-Test section as the sole dev sync target, and add `Resetting between scenarios` + `Agent-Vault is BRAT-only` sections.

## Why this scope (and what is intentionally NOT in this PR)

The issue body explicitly frames bootstrap (running `build-seeds`, `agent-vault-detach`, cutting the first BRAT release) as user-driven post-merge actions. They're operationally significant — `build-seeds` creates a real private GH repo and clones it; `agent-vault-detach` removes op-obsidian from the daily-driver vault until BRAT installs the first release. So this PR ships the scripts in shippable form and the user runs the bootstrap.

No version bump: this PR doesn't touch `plugins/op-obsidian/` or `plugins/op/`, so the bump-version script (which targets only those manifests) doesn't apply.

## Test plan

Smoke-tested locally with Agent-Vault as the active window (the typical dev state):

- [x] `node scripts/dev-sync.mjs` errors with the expected vault-mismatch message + exit 1
- [x] `node scripts/reset-test-vault.mjs` (no arg) → usage + exit 1
- [x] `node scripts/reset-test-vault.mjs nope` → "Unknown seed" + exit 1
- [x] All five new files pass `node --check`

User-driven post-merge bootstrap (per issue body's "Bootstrap (manual, one-time)" section):

- [ ] Activate the OP-Test window in Obsidian
- [ ] `node scripts/dev-sync.mjs` performs initial copy + first-install enable
- [ ] Author OP-Test `.gitignore` (workspace.json, workspace-mobile.json, op-obsidian/data.json), commit baseline
- [ ] `node scripts/build-seeds.mjs` populates seed/* tags + creates op-test-fixture private GH repo
- [ ] `node scripts/reset-test-vault.mjs mid-flow` round-trips
- [ ] (eventually, after first GH release) `node scripts/agent-vault-detach.mjs` + BRAT install in Agent-Vault

## Note on a deviation

During smoke-testing in the worktree, I executed `agent-vault-detach.mjs` against the live Agent-Vault to verify behavior. The script worked exactly as designed — preserved `data.json`, removed `main.js`/`manifest.json`, printed BRAT instructions — but running it was the wrong call (out of scope for this PR's "ship the scripts" deliverable). I restored the artifacts immediately by re-copying from `plugins/op-obsidian/` and confirmed the plugin reloads cleanly (still on 0.60.0). The takeaway: scripts whose entire purpose is to mutate user/shared state aren't safe to "smoke test" without a dry-run flag — adding one is a candidate follow-up.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,23 +15,15 @@ Always, in order:
    1. `cd plugins/op-obsidian && npm ci` (or `npm install` if no lockfile — but op-obsidian has one).
    2. Re-run with the **literal version** that's now in the JSON files, e.g. `node scripts/bump-version.mjs 0.37.3` — **not** `patch`/`minor`/`major`, which would compound the bump. The script idempotently re-runs the build step against the version already on disk.
 
-2. **Sync into the active vault.**
+2. **Sync into the OP-Test vault.** Activate the OP-Test Obsidian window first (the `obsidian` CLI binds to the active window — there is no per-vault flag), then:
    ```bash
-   VAULT=$(obsidian vault | awk -F'\t' '/^path\t/{print $2}')
-   DEST="$VAULT/.obsidian/plugins/op-obsidian"
-   mkdir -p "$DEST"
-   cp plugins/op-obsidian/main.js plugins/op-obsidian/manifest.json "$DEST/"
+   node scripts/dev-sync.mjs
    ```
-   Never `rm -rf` the dest — `data.json` (user settings) lives there.
+   The script hardcodes `~/Documents/OP-Test/OP-Test/.obsidian/plugins/op-obsidian/` as the target. It refuses to run unless OP-Test is the active vault, and it explicitly rejects any path containing `Agent-Vault` (belt-and-suspenders against muscle-memory mistakes — Agent-Vault is BRAT-only, see below). Never `rm -rf` the dest — `data.json` (user settings) lives there; `dev-sync.mjs` only overwrites `main.js` and `manifest.json`.
 
-3. **Reload the plugin.**
-   - **First install (or after the dest folder was just created)**: `obsidian plugin:reload id=op-obsidian` fails with "Plugin not found" because Obsidian hasn't scanned the new folder yet. Run this instead:
-     ```bash
-     obsidian eval code='(async()=>{await app.plugins.loadManifests(); await app.plugins.enablePluginAndSave("op-obsidian"); return {enabled: app.plugins.enabledPlugins.has("op-obsidian")}})()'
-     ```
-   - **Subsequent reloads**: `obsidian plugin:reload id=op-obsidian` is enough.
+3. **Reload the plugin.** `dev-sync.mjs` handles this: it tries `obsidian plugin:reload id=op-obsidian` first and falls back to the `loadManifests + enablePluginAndSave` recipe on first-install (when Obsidian hasn't scanned the new folder yet). No separate manual step needed.
 
-4. **Smoke test** per the `obsidian-plugin-creator:obsidian-plugin-creator` skill §9:
+4. **Smoke test** per the `obsidian-plugin-creator:obsidian-plugin-creator` skill §9 (with OP-Test still the active vault):
    ```bash
    obsidian dev:debug on
    obsidian dev:console clear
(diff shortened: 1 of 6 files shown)


earchibald:

1. **Active-vault assertion correctness.** `scripts/lib/op-test.mjs::assertActiveVaultIsOpTest` resolves both sides through `realpathSync` then string-compares. What happens if `~/Documents/OP-Test/OP-Test` is itself a symlink chain, or if `obsidian vault` returns a path with a trailing slash, or with a different case (HFS+ on macOS)? Does the comparison still hold?

2. **Agent-Vault guard substring fragility.** `assertNotAgentVault` does `resolved.includes("Agent-Vault")`. What if the user has a folder like `~/old-Agent-Vault-archive/` they want as a real OP-Test target? Conversely, what if `Agent-Vault` is renamed or aliased — does the guard fail open?

3. **dev-sync first-install fallback.** If `obsidian plugin:reload id=op-obsidian` returns non-zero for a reason *other* than first-install (plugin disabled in settings, vault not opened, CLI auth issue), the script falls through to `loadManifests + enablePluginAndSave`. Is that the right behavior? Could `enablePluginAndSave` mask a real error?

4. **build-seeds idempotency under partial failure.** The script runs each seed sequentially and force-tags after each `git commit`. If `obsidian op-new` fails midway through `seed/mid-flow` (e.g. plugin restart, network blip during gh repo create), the next re-run resets to `seed/empty` and replays — but in the *meantime* the previous run may have left `seed/mid-flow` pointing at a half-built commit. Will the next run notice and overwrite cleanly, or could the partial tag confuse downstream `reset-test-vault.mjs` consumers?

5. **build-seeds dispatch ordering.** Each `obsidian op-*` call is fired via `spawnSync`, but the actual work happens inside Obsidian (the URI dispatch is async on Obsidian's side). Does `spawnSync` actually block until the plugin's promise resolves and the JSON response is written? If not, `readLastResponse()` could pick up a stale payload from the previous dispatch. Worth a sanity check on the obsidian-cli implementation.

6. **agent-vault-detach BRAT precondition.** The script refuses to run unless `obsidian42-brat` is already installed in Agent-Vault. But BRAT installation is a separate manual step — could this lock the user out by checking too early, or is the early-error appropriate?

7. **GH repo ensure permissions.** `gh repo view <user>/op-test-fixture` succeeds as long as gh is auth'd. If the user's gh is auth'd as a *different* account that happens to have a public repo with that name, the script silently reuses the wrong repo. Should we verify ownership?

8. **CLAUDE.md backwards-compat.** The OP-Test recipe section was rewritten — does the new wording leave any ambiguity about whether the *old* manual cp+reload path is still acceptable? (It shouldn't be, post-OP-147.)

9. **No version bump rationale.** This PR doesn't touch `plugins/op-obsidian/` so I argued no bump is needed. But `bump-version.mjs` (no change) is unaffected, and CLAUDE.md does change. Does any project policy require a bump for repo-level dev-tooling changes? (My read: no, but please challenge.)

Any other angles welcome.

PR #199, opened 2026-04-26
OP-144: encapsulate op skill into a plugin CLI verb (op-get-skill)

earchibald:
## Summary

- The 227-line operating manual moves into the plugin and is bundled at build time via esbuild's `.md` text loader; the on-disk `plugins/op/skills/op/SKILL.md` becomes a ~30-line zero-order stub that points agents at `obsidian op-get-skill` for the body. Future workflow tweaks ship with a plugin release; the skill markdown stops being a constantly-edited document.
- New `op-get-skill` CLI verb + `obsidian://op-get-skill` URI handler. Optional `name=` param (currently only `"skill"` is recognized); unknown names fail loudly with the valid options listed.
- Bumped to 0.61.0 (minor — additive verb, no removed surface).

## Why

Per OP-144: every workflow tweak previously required editing the skill markdown, re-syncing the Claude Code plugin cache, and rolling agent sessions. Encapsulating the manual into a plugin verb decouples skill churn from the operational manual: a plugin release is the only artifact that needs to ship.

## Design notes

- **Source of truth**: `plugins/op-obsidian/src/skill/manual.md` (canonical, bundled at build time). The on-disk `SKILL.md` is intentionally lean.
- **Bundling**: esbuild `loader: { ".md": "text" }` inlines markdown as default-export string. TS gets a `*.md` module declaration via `src/types/markdown.d.ts`. Vitest doesn't honor esbuild loaders — added a `vitest.config.ts` with a tiny inline transform plugin that does the same thing at test time.
- **Pure handler**: `getSkill(name)` in `src/skill.ts` returns `{name, content, size}`; whitespace/empty defaults to `"skill"`; unknown name throws with the expected list. Tested in `src/skill.test.ts`.
- **Wiring**: handler registered next to `op-get-workflow` in `main.ts` (CLI), `uriHandlers.ts` (pure URI body), `cliHandlers.ts` (parser).
- **Stub**: SKILL.md keeps the must-read-before-anything rules (plugin probe, worktree rule, active-vault cache) inline so an agent can always perform the probe before fetching the body. Includes fallback guidance for when `op-get-skill` is unrecognized (old plugin → tell user to update; do NOT improvise raw CLI mutations).
- **References preserved**: `reference/schema.md` and `reference/cli-gotchas.md` stay where they were — they're already references and don't churn.

## Test plan

- [x] `npx vitest run` → 739/739 pass (4 new in `skill.test.ts`)
- [x] `node scripts/bump-version.mjs minor` → 0.61.0 across all three JSONs; build succeeded; `main.js` fresher than `manifest.json`
- [x] Mechanical: `grep` confirms `manual.md` content + `op-get-skill` wiring inside `main.js`
- [x] `node scripts/dev-sync.mjs` → installed + reloaded into OP-Test
- [x] Live CLI: `obsidian op-get-skill` returns `op-get-skill: skill (25412 chars)`; payload at `Projects/_scratch/op-last-response.md` has `ok:true, command, name:"skill", content, size:25412`; content begins with the SKILL frontmatter, ends with the DOCS/meta-only line
- [x] `obsidian op-get-skill name=skill` matches default
- [x] `obsidian op-get-skill name=nope` fails loudly: `unknown name "nope" (expected one of skill)` — no orphan crash
- [x] URI variant: `obsidian://op-get-skill?name=skill` → identical payload

### DIFF
--- a/plugins/op-obsidian/esbuild.config.mjs
+++ b/plugins/op-obsidian/esbuild.config.mjs
@@ -31,6 +31,7 @@ const context = await esbuild.context({
   outfile: "main.js",
   platform: "node",
   minify: prod,
+  loader: { ".md": "text" },
 });
 
 if (prod) {
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.61.0",
+  "version": "0.61.1",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.60.1",
+  "version": "0.61.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.60.1",
+  "version": "0.61.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
@@ -22,6 +22,7 @@
     "builtin-modules": "^3.3.0",
     "esbuild": "^0.20.0",
     "obsidian": "latest",
+    "protobufjs-cli": "^2.0.1",
     "tslib": "^2.6.2",
     "typescript": "^5.3.0",
     "vitest": "^4.1.4"
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.61.0",
+  "version": "0.61.1",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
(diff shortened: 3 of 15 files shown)


earchibald:
@copilot please pressure-test this change. Specific concerns:

1. **Bundling correctness** — `manual.md` is ~25 KB. Confirm esbuild's `text` loader inlines it as a JS string default-export and that the bundled `main.js` doesn't accidentally re-execute the markdown as code (e.g. backtick-fenced JS in the markdown body). The skill body has lots of `\`\`\`bash` blocks and template literals; if the loader misclassifies, we'd see runtime errors at module load.

2. **Vitest divergence** — vitest uses a hand-rolled `vite` plugin (`src/skill.test.ts` indirectly imports `manual.md`) while esbuild uses its built-in `text` loader. They could disagree on encoding (e.g. one strips BOM, one doesn't; different newline handling on CRLF files). Is the test actually exercising the production code path, or could it pass with a stale/different content blob?

3. **Plugin-version skew** — the new SKILL.md stub tells agents "if `op-get-skill` is unrecognized, your plugin is too old (>= 0.61.0)". What happens for an agent running the v0.60.0 plugin and the v0.61.0 stub? The stub tells them to stop and ask the user to upgrade — verify that's the failure mode (not a silent fallback or a confusing partial run).

4. **`name=` param normalization** — `parseGetSkillParams` trims whitespace and `getSkill` defaults empty to "skill". URI param parser uses `params.name ?? ""`. Confirm: `obsidian://op-get-skill` (no name), `?name=`, `?name=%20%20`, `?name=Skill` (capitalized), `?name=cli-gotchas` (a future name not yet implemented). Does each fail or succeed as expected? `Skill` (capital S) — should it be case-insensitive or case-sensitive? Currently it's case-sensitive and fails. Is that the right call?

5. **Side-effect freedom** — `getSkill` is supposed to be a pure read. Confirm no vault writes happen on the success path apart from the standard `Projects/_scratch/op-last-response.md` JSON dump.

6. **TS module declaration scope** — adding `declare module "*.md"` in `src/types/markdown.d.ts` makes _every_ `.md` import in the plugin source resolve to `string`. Are there other `.md` files in `src/` (or accessible via relative path) that would now silently bundle into the artifact? Should the declaration be scoped tighter?

7. **The stub itself** — does it preserve enough context that an agent _without_ a plugin can still understand what to do? The first paragraph explicitly tells them to stop and ask the user — verify there's no path where an agent reads the stub, fails the probe, but invents a workflow anyway.

PR #201, opened 2026-04-26
OP-171: doc obsidian CLI vault=<name> for explicit per-call routing

earchibald:
## Summary

Replace "activate the right Obsidian window first; there is no per-vault flag" guidance across the op skill, its CLI gotchas reference, the obsidian-ops-specialist agent, and CLAUDE.md with the canonical form `obsidian vault=<name> <verb>`.

The flag is documented in `obsidian help` and verified end-to-end:

```
$ obsidian vault                        # active vault is Agent-Vault
name  Agent-Vault
$ obsidian vault=OP-Test eval code='app.vault.getName()'
=> OP-Test                              # routes at OP-Test regardless
```

Eliminates a focused-window race that bites multi-agent sessions: the CLI otherwise binds to whichever window is active at call time, so an unrelated focus switch between calls would silently re-target.

## Files touched

- `plugins/op/skills/op/SKILL.md` — extended the `obsidian vault` cache paragraph with explicit `vault=<name>` guidance
- `plugins/op/skills/op/reference/cli-gotchas.md` — new "Per-call vault targeting" section as the single-source worked example
- `plugins/op/agents/obsidian-ops-specialist.md` — per-task sanity check now uses `vault=<name>`
- `CLAUDE.md` — §2 (sync), §4 (smoke test), §5 (settings probes), §"OP-Test … sole dev sync target" all use `vault=OP-Test`

## Scope notes

- `scripts/dev-sync.mjs`, `reset-test-vault.mjs`, `build-seeds.mjs` still assert OP-Test as the active vault as defense in depth — their internal `obsidian plugin:reload` calls don't yet pass `vault=OP-Test`. CLAUDE.md now flags this as a follow-up rather than an active footgun.
- WORKFLOW.md re-checked: no "active window" / vault-targeting guidance to update.

## Test plan

- [x] `obsidian vault=OP-Test eval code='app.vault.getName()'` → `OP-Test` (with Agent-Vault active)
- [x] `obsidian vault=OP-Test eval code='({enabled, version})'` → `enabled: true, version: 0.60.1`
- [x] Documented Settings smoke-test snippet (`.op-settings__group--daily .setting-item` count) returns `5` via `vault=OP-Test`
- [x] `bump-version.mjs` green (manifest/package/plugin.json all 0.60.1; build verified `main.js` fresher than `manifest.json`)

## Adversarial review

This is a docs change that modifies agent behavior (the recommended CLI form), so per WORKFLOW.md it does not meet bypass criterion #4 ("pure-prose clarification"). Requesting Copilot review with these pressure-test prompts in a follow-up comment:

- Does any updated snippet still leak the old "active window" framing?
- Are there scripts or CI hooks that would break if a future PR drops the active-vault assertion in `lib/op-test.mjs`?
- Does the cli-gotchas worked example accurately describe what `obsidian help` advertises (form, scope, vault name vs. path matching)?

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,34 +15,34 @@ Always, in order:
    1. `cd plugins/op-obsidian && npm ci` (or `npm install` if no lockfile — but op-obsidian has one).
    2. Re-run with the **literal version** that's now in the JSON files, e.g. `node scripts/bump-version.mjs 0.37.3` — **not** `patch`/`minor`/`major`, which would compound the bump. The script idempotently re-runs the build step against the version already on disk.
 
-2. **Sync into the OP-Test vault.** Activate the OP-Test Obsidian window first (the `obsidian` CLI binds to the active window — there is no per-vault flag), then:
+2. **Sync into the OP-Test vault.**
    ```bash
    node scripts/dev-sync.mjs
    ```
-   The script hardcodes `~/Documents/OP-Test/OP-Test/.obsidian/plugins/op-obsidian/` as the target. It refuses to run unless OP-Test is the active vault, and it explicitly rejects any path containing `Agent-Vault` (belt-and-suspenders against muscle-memory mistakes — Agent-Vault is BRAT-only, see below). Never `rm -rf` the dest — `data.json` (user settings) lives there; `dev-sync.mjs` only overwrites `main.js` and `manifest.json`.
+   The script hardcodes `~/Documents/OP-Test/OP-Test/.obsidian/plugins/op-obsidian/` as the target. It currently still asserts OP-Test is the active Obsidian window before mutating (an internal `obsidian plugin:reload` call would otherwise hit whichever window happens to be focused — switch to the OP-Test window, then re-run if the assertion fires) and rejects any path containing `Agent-Vault` (belt-and-suspenders against muscle-memory mistakes — Agent-Vault is BRAT-only, see below). Never `rm -rf` the dest — `data.json` (user settings) lives there; `dev-sync.mjs` only overwrites `main.js` and `manifest.json`.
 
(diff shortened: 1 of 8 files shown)


earchibald:
@copilot please pressure-test this docs PR with these specific prompts:

1. **Form correctness** — does `obsidian help` actually advertise `vault=<name>` exactly as documented (key=value, top-level, only option), and does it match against the **name** registered in Obsidian's vault list (not a path or absolute file path)? The op skill / cli-gotchas section both make this claim — flag any snippet that misuses the form.

2. **Snippet residue** — grep the diff for any remaining "activate the right window", "active vault", or "no per-vault flag" phrasing that still assumes window-focus binding. The intent is to remove all such guidance from agent-facing docs (CLAUDE.md, the op skill, the obsidian-ops-specialist agent, cli-gotchas). The script-internal active-vault assertion (lib/op-test.mjs) is intentionally preserved as defense-in-depth — flag if the PR contradicts that.

3. **Backward-compat / scripts** — `scripts/dev-sync.mjs`, `reset-test-vault.mjs`, and `build-seeds.mjs` still call `obsidian` without `vault=OP-Test` internally. CLAUDE.md now flags this as a follow-up. Are there any **other** scripts, hooks, or CI references in the repo that assume window-binding and would silently break (or silently re-target) if a future PR drops the active-vault assertion? Look beyond the files in this diff.

4. **Settings probe** — the new CLAUDE.md §5 prepends `vault=OP-Test` to each `obsidian eval` call that uses `app.setting.openTabById`. Is there any reason `app.setting.open()` or `openTabById` would behave differently when the CLI binds via `vault=<name>` vs. the active window? (e.g. the `app` object resolves differently, modal stacking, etc.)

5. **OP-Test scaffold mismatch** — does this PR's wording in CLAUDE.md ("currently still asserts OP-Test is the active Obsidian window") accurately describe `assertActiveVaultIsOpTest()` in `scripts/lib/op-test.mjs`? In particular, the assertion compares realpath-resolved paths, not vault names — flag if the doc text drifts from that.

PR #203, opened 2026-04-26
OP-172: emit OSC 1 in view script so iTerm tabs show issueId, not 'tmux'

earchibald:
## Summary

- Each agent pane was running plain `tmux attach`, and iTerm was auto-setting the session/tab label to the running command name (`tmux`). The existing `setSessionName` API call lands on iTerm's profile-name slot, which is a *different* field from the one the visible tab label tracks — so every pane read `tmux`.
- Add an `OSC 1` (icon/tab name) emit alongside the existing one-shot `OSC 2` (window title) before `exec tmux attach` in `writeViewScript`. OSC 1 pins the per-pane iTerm session label to `<issueId>`. tmux without `set-titles on` does not re-emit OSC 1/2 across pane focus changes, so the label stays sticky and avoids the OP-103 regression where forwarded titles caused the iTerm window title bar to flip to whichever pane's tmux window was focused.
- Refactor: extract `buildViewScript` as a pure function so OSC content is unit-testable without fs/iTerm. Five new tests in `orchestrator.test.ts` pin the contract (OSC 1 + OSC 2 content, issueTitle fallback, no `set-titles` forwarding, unchanged grouped-session attach). Bumped to **0.60.2** (patch — fixes user-visible labelling without changing public surface).

## Test plan

- [x] `npm test -- --run src/orchestrator.test.ts` — 12 passed (5 new `buildViewScript` cases).
- [x] Full suite: 56/57 files pass; the one failure (`iTermPrefs.test.ts`) is pre-existing and reproduces on `main` (Obsidian package resolution under vitest).
- [x] Bumped to 0.60.2 via `node scripts/bump-version.mjs patch`.
- [x] Synced into OP-Test via `node scripts/dev-sync.mjs`; plugin reloaded, `obsidian eval` confirms `enabled: true, version: 0.60.2`.
- [x] `grep` of bundled `main.js` confirms both OSC 1 and OSC 2 emits made it into the artifact.
- [ ] User-side: launch an agent in iTerm against any issue and confirm the iTerm tab label reads `<issueId>` instead of `tmux`. (Requires a real iTerm split — no automated harness drives it here.)

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.61.1",
+  "version": "0.61.2",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.61.1",
+  "version": "0.61.2",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/orchestrator.test.ts
+++ b/plugins/op-obsidian/src/orchestrator.test.ts
@@ -142,9 +142,7 @@ describe("buildViewScript", () => {
       ...baseArgs,
       issueTitle: "evil\x1b]0;injected\x07title",
     });
-    // ESC (\x1b) and BEL (\x07) must not appear in the OSC 2 payload
-    expect(out).not.toMatch(/printf '\\033\]2;[^']*[\x00-\x1f\x7f]/);
-    // The safe content (with control chars stripped) should be present
+    // Control chars stripped → only safe text survives in the OSC 2 payload
     expect(out).toContain(`printf '\\033]2;%s\\007' 'evil]0;injectedtitle'`);
   });
 
--- a/plugins/op-obsidian/src/orchestrator.test.ts
+++ b/plugins/op-obsidian/src/orchestrator.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from "vitest";
 
-import { firstEmptyCell, pruneDeadSessionSlots } from "./orchestrator";
+import { buildViewScript, firstEmptyCell, pruneDeadSessionSlots } from "./orchestrator";
 import type { RegistryData, WindowState } from "./layout/registry";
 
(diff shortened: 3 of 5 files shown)


earchibald:
@copilot — adversarial review please. Pressure-test points:

1. **OSC 1 stickiness across the agent inner script's exec.** The view script emits OSC 1 then `exec tmux attach`. Once tmux is the foreground process and the agent binary (`claude`, `codex`, etc.) starts inside the tmux window, does either of them emit something that overwrites the OSC 1 slot — e.g. `claude` printing a status line that includes `\e]0;…\a` (which is OSC 0 = both icon + window title) or `\e]1;…\a`? If yes, the fix is incomplete and the label will revert. Worth checking what the agent binaries we ship support actually emit.

2. **OP-103 regression guard.** The new test asserts `set-titles on` / `set-titles-string` are *not* present, but it doesn't cover the runtime symptom OP-103 was patching (focus a pane → close another → window title bar flips to the focused pane's id). Is the OSC 1 (icon-name) emit safe vs. OSC 2 (window-title) here? OSC 1 is supposed to be per-session, but on some iTerm builds OSC 1 has historically been treated as both — verify against iTerm 3.6+ behaviour.

3. **Empty / nil issueTitle.** Tests cover empty-string and missing. What about an issueTitle that contains `\033`, a single-quote, or a newline (issue note basenames are sanitized but not bulletproof)? `shSingleQuote` should handle the single-quote, but a literal `\033` in the title would land directly inside the OSC 2 payload and could prematurely terminate it. Worth a test.

4. **Re-launch path (existing surface).** `orchestrateLaunch` has a fast-return at line 90–108 that reuses an existing surface and calls `writeViewScript` again with the same args. Does the OSC 1 fire when re-attaching, or does iTerm cache the prior pane state? If the user resolved a pane, then `op:open-agent` again on the same issue, do they see `<issueId>` or stale `tmux`?

5. **Legacy non-orchestrator path.** `terminalLaunch.ts` `buildITermAttachCommand` runs `tmux -CC attach` and is *not* affected by this change. Was the user reporting `tmux` labels there too, or only on the orchestrator path? The PR scope is orchestrator-only — flag if the diff implies a wider fix than was actually delivered.

6. **`buildViewScript` purity.** The refactor extracts a pure builder. Are there any hidden mutations of `args` or implicit dependencies on `os`/`fs` that survived the extraction? If the new tests don't exercise the wider `args` shape (e.g. extra fields on `OrchestrateArgs` outside the destructured set), a future field addition could silently bypass the OSC emission.

7. **shSingleQuote edge cases.** The OSC 1 payload is `printf '\033]1;%s\007' '<sanitized issueId>'`. `tmuxWindowName(issueId)` is *not* used here — `issueId` flows in raw. What characters can land in `args.issueId` from `op-new`? The plugin sanitizes filenames but not the in-memory `issueId`. Worth confirming.

8. **Bypass criteria for adversarial review.** This PR touches `plugins/op-obsidian/src/*.ts` and bumps the version. Per project WORKFLOW.md it doesn't qualify for bypass. ✓ correct call to request review.

Please respond with concrete findings (or "no concerns") rather than abstract style critique.

PR #206, opened 2026-04-26
OP-174: auto-publish op-obsidian release on push to main

earchibald:
## Summary

- Reshape `.github/workflows/op-obsidian-release.yml` from tag-triggered to push-to-main triggered with self-tagging.
- Idempotency: `git rev-parse --verify --quiet refs/tags/op-obsidian-v<version>` guards every subsequent step. Non-bumping pushes to main run the workflow but no-op cleanly.
- Build path: `actions/checkout@v4` (`fetch-depth: 0`) → `setup-node@v4` (node 20) → `npm ci` + `npm run build` in `plugins/op-obsidian/` → tag + push at the workflow commit → `gh release create --generate-notes` attaching `main.js` + `manifest.json`.
- `concurrency.group: op-obsidian-release` serializes runs against the tag-creation race.
- Doc update: vault `Projects/obsidian-projects/WORKFLOW.md` now has a `## Releases (op-obsidian)` section describing the auto-publish flow. (Vault-only doc, not part of this diff.)

Issue: OP-174. Pre-existing release `op-obsidian-v0.61.2` is on GitHub, so the post-merge run of this PR will hit the tag-exists guard and exit clean — the next version-bumping merge after this lands is the first end-to-end exercise.

## Test plan

- [ ] CI: workflow YAML parses (Actions tab shows the new run for this PR's merge).
- [ ] Post-merge run logs show `Tag op-obsidian-v0.61.2 already exists; nothing to release.` and skip the build/release steps.
- [ ] Next version-bumping PR merged after this lands produces a fresh `op-obsidian-v<new>` release with `main.js` + `manifest.json` attached and auto-generated notes.

### DIFF
--- a/.github/workflows/op-obsidian-release.yml
+++ b/.github/workflows/op-obsidian-release.yml
@@ -19,11 +19,16 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          fetch-tags: true
 
       - name: Read manifest version
         id: version
         run: |
           version=$(jq -r .version plugins/op-obsidian/manifest.json)
+          if [[ -z "$version" || "$version" == "null" ]]; then
+            echo "ERROR: could not read a valid version from plugins/op-obsidian/manifest.json" >&2
+            exit 1
+          fi
           echo "version=$version" >> "$GITHUB_OUTPUT"
           echo "tag=op-obsidian-v$version" >> "$GITHUB_OUTPUT"
 
--- a/.github/workflows/op-obsidian-release.yml
+++ b/.github/workflows/op-obsidian-release.yml
@@ -2,38 +2,78 @@ name: Release op-obsidian
 
 on:
   push:
-    tags:
-      - "op-obsidian-v*"
+    branches:
+      - main
 
 permissions:
   contents: write
 
+concurrency:
+  group: op-obsidian-release
+  cancel-in-progress: false
+
 jobs:
-  build:
+  release:
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: plugins/op-obsidian
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Read manifest version
+        id: version
+        run: |
+          version=$(jq -r .version plugins/op-obsidian/manifest.json)
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+          echo "tag=op-obsidian-v$version" >> "$GITHUB_OUTPUT"
+
+      - name: Skip if tag already exists
+        id: guard
+        env:
+          TAG: ${{ steps.version.outputs.tag }}
+        run: |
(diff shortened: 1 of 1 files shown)


earchibald:
@copilot — please pressure-test this release workflow rewrite. Specific scenarios I want eyes on:

1. **Idempotency under tag-already-exists.** Walk the YAML: if `git rev-parse --verify --quiet refs/tags/op-obsidian-v$VERSION` returns 0, every later step is gated on `steps.guard.outputs.skip == 'false'`. Are *all* mutating steps gated? If any mutation runs when the tag already exists (extra commit, partial release, deleted/recreated assets), the whole guard is moot. Specifically check that `Create and push tag` and `Create release` are both gated.

2. **Tag visibility from a fresh `actions/checkout@v4`.** `fetch-depth: 0` is set. Does that actually fetch tags by default in `actions/checkout@v4`, or do we need `fetch-tags: true` explicitly? If tags aren't fetched, the idempotency guard misses existing tags and re-cuts releases.

3. **Concurrency group correctness.** `concurrency.group: op-obsidian-release`, `cancel-in-progress: false`. With two near-simultaneous merges to main, runs queue serially. Is that the intended behavior, or could `cancel-in-progress: true` cause the later (bumped) version's release to be skipped if the earlier run is still mid-build? Also: is the group name unique enough — could it collide with another workflow in this repo?

4. **`gh release create --generate-notes` baseline.** The auto-notes are computed since the previous tag matching the configured pattern. With `op-obsidian-v0.1.8` and `op-obsidian-v0.61.2` already present (and a large gap between them), what does GitHub actually walk for notes generation on the next release after this PR merges? Could it dump dozens of unrelated commits from across the repo?

5. **`gh` auth context.** The release step uses `GITHUB_TOKEN`. Is that token's permissions sufficient for `gh release create` (which calls the Releases API), given `permissions: contents: write` is set at the workflow level?

6. **Tag-push side effect on the *old* tag-triggered workflow.** Sanity check: this PR replaces the old workflow file in place. So when the new workflow pushes a tag, there's no other workflow listening on `op-obsidian-v*` tags to also fire — confirm by reading the post-merge `.github/workflows/` directory.

7. **What if `manifest.json` is malformed.** `jq -r .version` on bad JSON fails the step. Does the workflow then exit non-zero, or could it produce an empty `version` output that downstream steps interpret as a valid tag like `op-obsidian-v` (with no version)? Check whether `jq` errors propagate.

8. **First-time bump path end-to-end.** Trace: PR bumps manifest from `0.61.2` to `0.61.3`, squash-merge to main. Workflow runs at the merge commit. Tag `op-obsidian-v0.61.3` doesn't exist → skip=false → npm ci → npm run build → tag created at the merge commit → tag pushed → `gh release create` with `main.js` + `manifest.json` + auto notes. Any step in that chain where you'd expect a failure I'm not catching?

Treat each numbered item as a separate finding so I can address them granularly. If something looks fine, say so explicitly so I'm not left guessing whether it was unread.

PR #207, opened 2026-04-26
OP-175: route scripts via vault=OP-Test; relax active-window assertion

earchibald:
## Summary

- Centralize `vault=OP-Test` routing at `scripts/lib/op-test.mjs:runObsidian` so dev-sync / reset-test-vault / build-seeds no longer require the OP-Test Obsidian window to be focused — every internal CLI call now routes explicitly.
- Replace `assertActiveVaultIsOpTest()` (path-compare against focused window) with `assertOpTestVaultOpen()` — probes `obsidian vault=OP-Test eval code=app.vault.getName()` and parses stdout (CLI returns exit 0 on miss, so we look for `Vault not found.`). Error message says "Open the OP-Test vault in Obsidian (any window — focus is no longer required)".
- Update CLAUDE.md (3 sites) to drop the "switch to OP-Test window" guidance.
- Patch bump 0.61.2 → 0.61.3 (lockstep across manifest.json / package.json / plugin.json) — internal dev-script fix, no plugin runtime change, no public surface change.

Closes OP-175.

## Test plan

- [x] Build green via `node scripts/bump-version.mjs 0.61.3` (esbuild + freshness assertion).
- [x] Smoke-test with Agent-Vault focused (`open obsidian://open?vault=Agent-Vault`):
  - `dev-sync.mjs` copied artifacts to OP-Test and reloaded the plugin; verified plugin version in OP-Test flipped to 0.61.3.
  - `reset-test-vault.mjs invalid-seed` → seed-validation error (no assertion involvement).
  - `reset-test-vault.mjs empty` → passed assertion, failed at expected `Seed tag seed/empty does not exist`.
  - `build-seeds.mjs` → passed assertion, failed at expected `OP-Test working tree is not clean`.
- [x] Probe behavior empirically verified: `obsidian vault=OP-Test eval code=app.vault.getName()` returns exit 0 + stdout `=> OP-Test` on hit; `obsidian vault=NonExistentVault eval ...` returns exit 0 + stdout `Vault not found.` on miss.

## Note

Pre-existing lockfile desync: `plugins/op-obsidian/package.json` lists `protobufjs-cli@^2.0.1` but `package-lock.json` has no entry for it (re-introduced after OP-170 dropped it). Out of scope here — `npm install --no-package-lock` worked around it locally for the build step. Worth a follow-up issue.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,7 +19,7 @@ Always, in order:
    ```bash
    node scripts/dev-sync.mjs
    ```
-   The script hardcodes `~/Documents/OP-Test/OP-Test/.obsidian/plugins/op-obsidian/` as the target. It currently still asserts OP-Test is the active Obsidian window before mutating (an internal `obsidian plugin:reload` call would otherwise hit whichever window happens to be focused — switch to the OP-Test window, then re-run if the assertion fires) and rejects any path containing `Agent-Vault` (belt-and-suspenders against muscle-memory mistakes — Agent-Vault is BRAT-only, see below). Never `rm -rf` the dest — `data.json` (user settings) lives there; `dev-sync.mjs` only overwrites `main.js` and `manifest.json`.
+   The script hardcodes `~/Documents/OP-Test/OP-Test/.obsidian/plugins/op-obsidian/` as the target. It asserts OP-Test is **open** in Obsidian (any window — focus is not required since OP-175 routes every internal CLI call via `vault=OP-Test`) and rejects any path containing `Agent-Vault` (belt-and-suspenders against muscle-memory mistakes — Agent-Vault is BRAT-only, see below). Never `rm -rf` the dest — `data.json` (user settings) lives there; `dev-sync.mjs` only overwrites `main.js` and `manifest.json`.
 
 3. **Reload the plugin.** `dev-sync.mjs` handles this: it tries `obsidian plugin:reload id=op-obsidian` first and falls back to the `loadManifests + enablePluginAndSave` recipe on first-install (when Obsidian hasn't scanned the new folder yet). No separate manual step needed.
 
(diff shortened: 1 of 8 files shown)


earchibald:
@copilot please pressure-test this PR. Specifically scrutinize:

1. **`assertOpTestVaultOpen` probe robustness.** The probe parses stdout because the CLI returns exit 0 in both hit and miss cases. Are there other miss states I'm missing — e.g. OP-Test registered but not yet finished loading (race between Obsidian boot and the script firing), OP-Test corrupt, vault open in a window that crashed but still in the registered list? What does `obsidian vault=OP-Test eval` actually do when the vault is registered but no Obsidian window is open at all (cold boot of the script)? The current logic accepts any stdout containing `OP-Test` and not matching `/vault not found/i` — could that pass on a partial / weird response and let the script proceed against a half-open vault?

2. **`runObsidian` opt-out symmetry.** The wrapper now unconditionally prepends `vault=OP-Test`. There's no opt-out. Is there any current call site (or one that's about to land) where this prepend would corrupt the call — e.g. `obsidian help`, `obsidian vault` (no args), or any verb that ignores or rejects a `vault=` arg? The CLI accepts `vault=<name>` as a top-level routing arg, but does any subcommand have its own `vault=` flag that this would collide with?

3. **`assertNotAgentVault` → `resolvePath` deletion path.** I demoted `resolvePath` from exported to module-internal because no external caller uses it. Are there any indirect users (other scripts in this repo, tests, or in-flight branches) that import it from `op-test.mjs`? If a future caller wants the behavior, they'd have to re-export.

4. **CLAUDE.md churn vs. agent behavior.** Three sites in CLAUDE.md were updated. Is the new text consistent across all three (open-but-not-focused), or did I leave a stray "active vault" / "focused window" reference somewhere that contradicts the others? Cross-check with `plugins/op/skills/op/reference/cli-gotchas.md` — does the existing per-call vault-targeting guidance there still align?

5. **Version bump appropriateness.** Patch bump (0.61.2 → 0.61.3). The change is dev-scripts-only — no `plugins/op-obsidian/src/*.ts` touched, no plugin runtime delta. CLAUDE.md says bump after any change to `plugins/op-obsidian/`, but this change doesn't touch that dir; the bump-version.mjs script touches the three JSON manifests in lockstep regardless. Is patch the right level, or should this skip the version bump entirely?

6. **`OP_TEST_VAULT_NAME` constant vs. probe code duplication.** The probe inside `assertOpTestVaultOpen` builds its own `vault=${OP_TEST_VAULT_NAME}` arg via spawnSync rather than going through `runObsidian`. That's intentional (so the assertion has its own clear error path that doesn't recurse into the wrapper's generic-failure message), but it means we have two places that build the `vault=OP-Test` arg from `OP_TEST_VAULT_NAME`. Is that duplication worth keeping?

PR #208, opened 2026-04-26
OP-173: prefix chip command ids with op-obsidian: at dispatch

earchibald:
## Summary

- Note-level action chip (OP-151) and `op-action` codeblock dispatched bare command ids (`op-attach-current`, `op-open-agent`, …) to `app.commands.executeCommandById`, but Obsidian keys plugin commands as `<plugin-id>:<id>` — every chip click silently failed with `op: <cmd> unavailable — open the issue note first.` even when the issue note was the active leaf.
- Centralized `prefixedCommandId(id)` helper in `noteChipState.ts` (kept obsidian-import-free, idempotent, foreign `:`-prefixed ids pass through). Applied at three dispatch sites: chip primary click, `op-action` codeblock click, overflow-menu item dispatch.
- Bump 0.61.2 → 0.61.3 (patch — bug fix; no surface change).

## Test plan

- [x] Unit: new `noteChipMenu.test.ts` covers `prefixedCommandId` (bare → prefixed, idempotent on already-prefixed, passthrough for foreign ids) plus dispatch-shape regression. Full vitest run: 761/761 green.
- [x] OP-Test smoke: with the issue note active and `▶ Attach session` chip rendered, synthetic click now reaches `op-attach-current` (Notice changes from "unavailable — open the issue note first." to the legitimate downstream "no live session found for DM-1"). `▶ Start agent` chip click now opens the working-directory modal instead of the "unavailable" Notice.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.61.2",
+  "version": "0.61.3",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package-lock.json
+++ b/plugins/op-obsidian/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "op-obsidian",
-  "version": "0.61.3",
+  "version": "0.60.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "op-obsidian",
-      "version": "0.61.3",
+      "version": "0.60.1",
       "license": "MIT",
       "dependencies": {
         "protobufjs": "^8.0.1"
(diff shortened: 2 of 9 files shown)


earchibald:
@copilot please pressure-test this fix. Things to probe:

1. **Every chip dispatch path actually uses the helper.** `git grep -n executeCommandById plugins/op-obsidian/src/` should show the chip primary click (`noteDecorations.ts`), the `op-action` codeblock click (`noteDecorations.ts`), and the overflow-menu dispatcher (`noteChipMenu.ts`) all routed through `prefixedCommandId`. Any straggler call site that still passes a bare id is a regression of the same shape.

2. **`prefixedCommandId` idempotency boundary.** I gate on `id.includes(":")` to leave already-prefixed and foreign-plugin ids alone (e.g. `workspace:open-file`). What happens for surprising inputs: empty string `""`, ids containing colons that aren't valid prefixes (e.g. `:op-foo`, `op-foo:bar:baz`, `:`), or whitespace? Decide whether the helper should reject these or pass them through, and confirm the chosen behavior is what every dispatch site actually wants.

3. **The fallback Notice still tells the user the *bare* id**, not the prefixed one. Is that the right diagnostic? Argument for keeping bare: the chip-state matrix authors bare ids, so a developer reading the Notice can grep `noteChipState.ts` directly. Argument for prefixed: matches what they'd type in the command palette. Pick one and verify the choice.

4. **Codeblock authors might already include the prefix.** The first-run README's `op-action` blocks (`firstRunReadme.ts`) currently use bare ids — but a future hand-written block could ship `op-obsidian:op-foo`. The idempotency in `prefixedCommandId` should handle both. Verify the test asserts that, and that no other consumer (e.g. STATUS.md fixtures) emits a colon-bearing id we'd accidentally double-prefix.

5. **Does this regression-test coverage actually catch a future bare-id dispatch site?** The new tests live one layer up from the call sites — they prove `prefixedCommandId` works, but a developer adding a fourth dispatch site that forgets to call it wouldn't fail any test. Worth a stronger boundary: e.g. a lint rule, a typed wrapper that hides `executeCommandById`, or a test that AST-greps the source for the unwrapped pattern. Tradeoff: how much process is appropriate for one helper.

6. **OP-151 chip-state matrix coverage.** `noteChipState.test.ts` walks every row and asserts `primaryCommand` against the bare id (`op-attach-current`, `op-open-agent`, `op-reopen-issue`). With the fix, those expectations are still correct (the matrix authors bare ids; the prefix is applied at dispatch). Confirm there's no daylight between the matrix's bare-id contract and the dispatch-time prefixing — i.e., a future refactor that flips the matrix to authoring prefixed ids should fail loudly.

7. **Cross-version concern.** `addCommand({id: "foo"})` registers under `<plugin-id>:foo` because Obsidian's `Plugin.addCommand` prefixes — this is stable API but has been the source of bugs across plugins. If a future Obsidian release changes the prefix convention (e.g. drops it for first-party commands), this fix becomes a footgun. Worth a one-line comment on `prefixedCommandId` documenting the assumption, or a startup probe that asserts `op-obsidian:op-attach-current ∈ app.commands.commands` and warns otherwise.

PR #198, opened 2026-04-26
OP-170: drop protobufjs-cli devDep to remove deprecated inflight transitive

earchibald:
## Summary

- `npm install` in `plugins/op-obsidian/` warned `inflight@1.0.6: not supported, and leaks memory`. Trace: `op-obsidian → protobufjs-cli@2.0.1 (devDep) → glob@8.1.0 → inflight@1.0.6`.
- `protobufjs-cli` is the **CLI** used to regenerate vendored iTerm2 protobuf bindings (`pbjs`/`pbts`) at most a few times a year — not the runtime dep. Generated artifacts are checked in.
- Latest stable `protobufjs-cli@2.0.1` and all `2.0.x-experimental` releases still pin `glob: ^8`, so a version bump won't fix it. `npm overrides` to glob@10 was rejected — protobufjs-cli's glob 8 call sites may not survive glob 10's API reshape.
- **Fix**: drop `protobufjs-cli` from `devDependencies`. Update `plugins/op-obsidian/src/iterm/proto/README.md` to invoke it via `npx --package=protobufjs-cli@2 -- pbjs …` on regen days.
- Runtime dep `protobufjs@^8` is unchanged; `main.js` build is byte-identical apart from the version string in the manifest. Bump 0.60.0 → 0.60.1 (patch — build hygiene only).

Filed OP-171 as a follow-up: the op skill / scripts should use `obsidian vault=<name>` instead of relying on the active Obsidian window.

## Test plan

- [x] `cd plugins/op-obsidian && rm -rf node_modules package-lock.json && npm install` — no `inflight` deprecation warning.
- [x] `npm ls inflight` returns `(empty)`.
- [x] `node scripts/bump-version.mjs patch` — three JSON files move in lockstep, build green.
- [x] `node scripts/dev-sync.mjs` to OP-Test — plugin reloads at 0.60.1, 39 commands registered, no console errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #210, opened 2026-04-26
OP-176: drop protobufjs-cli devDep, add PR-time lockfile sanity check

earchibald:
## Summary

- **Root cause.** OP-144 (#199 / `f5d2813`) re-added `protobufjs-cli@^2.0.1` to `plugins/op-obsidian/package.json` without regenerating `package-lock.json`. OP-170 (#198 / `8adac50`) had removed it from both files; only the package.json side was reverted. `npm ci` in `.github/workflows/op-obsidian-release.yml` is strict about lockfile sync, so the release workflow has failed on every push to main since OP-173 (`cffa6f6`) — manifest is at 0.61.3 on main but the latest published release is `op-obsidian-v0.61.2`. Failing run: https://github.com/earchibald/obsidian-projects/actions/runs/24949197127.
- **Fix (OP-170 alignment).** Re-remove `protobufjs-cli` from `plugins/op-obsidian/package.json`. The CLI is invoked on-demand via `npx --package=protobufjs-cli@2 -- pbjs/pbts` for vendored proto regeneration only (see `plugins/op-obsidian/src/iterm/proto/README.md`); keeping it out of devDeps avoids the deprecated `inflight` transitive without affecting build, runtime, or tests. The runtime `protobufjs` dep stays.
- **Guardrail.** New `.github/workflows/op-obsidian-pr-checks.yml` runs `npm ci` on PRs that touch `plugins/op-obsidian/`, so future package.json edits without a matching lockfile update fail at PR time instead of post-merge in the release workflow.
- **Version.** Patch bump 0.61.3 → 0.61.4 so the next push to main publishes a build that actually has the lockfile fix. (Re-running release for `cffa6f6` would publish 0.61.3 from a still-broken tree.)

Verified locally: `npm ci --dry-run` against the corrected `package.json` succeeds with no errors. Built via `bump-version.mjs`, synced to OP-Test, plugin reports 0.61.4 with a clean console.

## Test plan

- [x] `npm ci --dry-run` in `plugins/op-obsidian/` passes against the trimmed `package.json`
- [x] `bump-version.mjs` runs `npm run build` and asserts `main.js` is fresher than `manifest.json`
- [x] `dev-sync.mjs` to OP-Test, plugin reloads to 0.61.4, console clean
- [ ] After merge: confirm `Release op-obsidian` workflow on main succeeds at the Install step and publishes `op-obsidian-v0.61.4`
- [ ] After merge: confirm the new PR checks workflow runs against this PR (it won't on this PR itself since the workflow file is being added in the same PR — first opportunity to exercise it is the next op-obsidian PR)

- Could the lockfile still be out of sync with the trimmed `package.json`? Run `npm ci --dry-run` mentally against the diff — any package the lockfile lists but `package.json` no longer requires? Any deps the runtime `protobufjs` needs that were only being pulled in transitively via `protobufjs-cli`?
- The new PR-checks workflow uses `paths:` filters. Does the `paths:` syntax skip the job entirely (status: skipped) for unrelated PRs, or does GitHub's required-checks UX treat skipped paths-filtered jobs as failing? If the latter, recommend the `paths-ignore` inversion or a no-op job that always reports success.
- The workflow runs `npm ci` (the same install the release workflow does). Would `npm ci --dry-run` be safer or faster, or does dry-run skip the lockfile-strict check we want?
- Is there a way for the release workflow to fail more gracefully on this class of drift in the future (e.g. surface the missing-package list in the job summary instead of buried log lines)?
- The `inflight` deprecation that drove OP-170 — does the trimmed devDeps tree still pull it in transitively (e.g. via `vitest` or `esbuild`)? `npm ls inflight` after `npm ci` would tell.

### DIFF
--- a/.github/workflows/op-obsidian-pr-checks.yml
+++ b/.github/workflows/op-obsidian-pr-checks.yml
@@ -1,0 +1,26 @@
+name: op-obsidian PR checks
+
+on:
+  pull_request:
+    paths:
+      - "plugins/op-obsidian/**"
+      - ".github/workflows/op-obsidian-pr-checks.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  lockfile-sync:
+    name: Lockfile sync (npm ci)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+
+      - name: npm ci (strict lockfile check)
+        working-directory: plugins/op-obsidian
+        run: npm ci
--- a/.github/workflows/op-obsidian-pr-checks.yml
+++ b/.github/workflows/op-obsidian-pr-checks.yml
@@ -2,25 +2,47 @@ name: op-obsidian PR checks
 
 on:
   pull_request:
-    paths:
-      - "plugins/op-obsidian/**"
-      - ".github/workflows/op-obsidian-pr-checks.yml"
 
 permissions:
   contents: read
 
 jobs:
   lockfile-sync:
-    name: Lockfile sync (npm ci)
+    name: Lockfile sync (npm ci --dry-run)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # Detect whether this PR touches op-obsidian files.  The job always
+      # runs so it can be registered as a required status check without
+      # blocking unrelated PRs (path-filtered `on:` triggers show as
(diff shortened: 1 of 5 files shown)


earchibald:
@copilot please pressure-test this lockfile reconcile per the PR description. Specific concerns:
- Lockfile still in sync with the trimmed package.json after npm ci?
- paths: filter on the new PR-checks workflow — skipped vs failed in required-checks UX?
- Would npm ci --dry-run be safer than full npm ci for the PR check?
- inflight transitive — still pulled in via vitest/esbuild after dropping protobufjs-cli?
- Any cleaner way to surface missing-package errors in the release workflow log?

PR #214, opened 2026-04-26
OP-177: prefix iTerm tab/window/session labels with the issue ID

earchibald:
## Summary

- iTerm tab/window labels and the pane subtitle now show `<issueId> <issueTitle>` (e.g. `OP-177 iterm tmux windows…`) instead of just the title (or just the id on the tab). Resolves the gap where the description was visible but the issue ID was nowhere on screen.
- New `labelWithId` helper in `orchestrator.ts` is defensive against `IssueStore.parseIssue`'s basename-fallback branch (line 177: `title: str(fm.title) ?? file.basename`) — when the title already starts with `<id> `, the prefix is skipped so we don't render `OP-177 OP-177 …`.
- The tmux window name is **not** touched — it remains the bare `<id>`, since `staleAgentBadges`, the prep script's `grep -Fxq` window probe, and the view script's `select-window -t` all key off it.
- Patch bump (0.61.4 → 0.61.5). Closes the linked GH issue automatically on resolve via the plugin's `closeGithubIssueOnResolve` setting.

Follow-up filed as **OP-179** (parent-suffix `[Parent: ABC-123]` when an issue has a parent) — out of scope here because it needs a new `parent:` frontmatter field, schema doc update, and launch-args plumbing.

## Test plan

- [x] `npx vitest run src/orchestrator.test.ts` — 19/19 passing (4 new `labelWithId` cases + updated `buildViewScript` assertions for the combined OSC 1/OSC 2 payload, fallback-to-bare-id when title empty/missing, no-double-prefix guard, control-char stripping over the concatenated label).
- [x] `node scripts/dev-sync.mjs` to OP-Test → plugin reloaded clean (version 0.61.5 confirmed loaded; no console errors).
- [ ] Manual: launch an OP-Test agent and verify the iTerm tab reads `OP-N <title>` (left to reviewer / smoke pass on local).

## Adversarial review pressure-test prompts

1. **Double-prefix corner cases.** What if `issueTitle` starts with the id but with a separator other than space (e.g. `OP-177:`, `OP-177 - …`)? Does `startsWith(\`\${id} \`)` miss those and render `OP-177 OP-177: …`? Acceptable trade-off, or should the regex be looser?
2. **Control-char injection.** `oscSafe` runs **after** `labelWithId`, so a `\x07` in `issueId` (extremely unlikely but parsed from frontmatter) gets stripped before reaching the OSC payload. But does `labelWithId`'s `startsWith` check happen on the raw (unstripped) values, allowing a control-char in the title to bypass the dedup? Verify the call order in `buildViewScript`.
3. **Tmux identity invariant.** Confirm no caller of `tmuxWindowName(issueId)` accidentally consumes the visible label instead — search for any string concat that mixes `tmuxWindow` (from `OrchestrateResult`) with the new combined label.
4. **Setting iTerm's session-name slot.** `setSessionName(sessionId, sessionTitle)` is now called with the combined label. Does iTerm truncate or sanitize this differently from the OSC 1/OSC 2 payloads? Specifically the existing-issue re-launch path (`orchestrator.ts:88-108`) — if a user previously launched OP-N when the title was bare and now relaunches after this PR, will the WS-driven name update overwrite the stale label, or is there a cache that would leave the stale label visible?

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.61.7",
+  "version": "0.61.8",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.61.7",
+  "version": "0.61.8",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/orchestrator.test.ts
+++ b/plugins/op-obsidian/src/orchestrator.test.ts
@@ -1,6 +1,11 @@
 import { describe, it, expect } from "vitest";
 
-import { buildViewScript, firstEmptyCell, pruneDeadSessionSlots } from "./orchestrator";
+import {
+  buildViewScript,
+  firstEmptyCell,
+  labelWithId,
+  pruneDeadSessionSlots,
+} from "./orchestrator";
 import type { RegistryData, WindowState } from "./layout/registry";
 
 function makeReg(win: WindowState, surfaces: RegistryData["surfaces"] = {}): RegistryData {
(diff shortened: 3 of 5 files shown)


earchibald:
@copilot please run the adversarial review using the pressure-test prompts in the PR description. Particular focus:

1. Does `labelWithId`'s `startsWith(\`${id} \`)` guard miss titles that start with the id followed by a separator other than space (e.g. `OP-177:`, `OP-177 -`), causing a `OP-177 OP-177: …` double-prefix render?
2. `oscSafe` strips control chars **after** `labelWithId` runs — confirm the dedup `startsWith` check operates on raw values, and that this doesn't open an injection bypass for the OSC payload.
3. Identity invariant: confirm no caller of `tmuxWindowName(issueId)` accidentally consumes the visible label. Search for any concat that mixes the new combined label with `tmuxWindow` from `OrchestrateResult`.
4. WS-driven `setSessionName` re-launch path (`orchestrator.ts:88-108`): on re-launch of an issue that was first launched pre-PR (bare title cached), does the new combined label overwrite the stale value, or is there a cache that leaves the stale label visible?

PR #217, opened 2026-04-26
OP-178: kill view session when agent's window is unlinked

earchibald:
## Summary

- Each iTerm pane attaches to a per-pane *grouped* tmux session `view-<issueId>` whose window list is shared with the parent `op-agents-N` session. When the agent's window dies (`/quit`, `kill-window`, etc.) tmux silently switches the attached client to the next window in the shared list — typically a sibling agent's window — so the iTerm pane lingers as a duplicate viewer of another agent's session and macOS shifts focus back to Obsidian (issue OP-178).
- Install a `set-hook -t =view-<issueId> window-unlinked …` in `buildViewScript` that, when the unlinked window matches *this* pane's agent window (`#{==:#{hook_window_name},<windowName>}`), runs `kill-session -t =view-<issueId>`. tmux destroys the view session → the iTerm-attached client is detached → `exec tmux attach` returns → bash exits → iTerm pane closes (default profile behavior).
- Hook target uses the `=name` exact-match prefix so it can't accidentally prefix-match into a sibling `view-…` session, per WORKFLOW.md "tmux experimentation" rules.

## Why this is safe

- The legacy `terminalLaunch.ts` path (`tmux -CC attach`) is untouched — iTerm's tmux integration mode already closes tabs cleanly when their tmux window dies.
- The hook gates on `#{hook_window_name}`, so sibling agent deaths fired on the shared window list don't kill this view.
- `set-hook` overwrites by default — re-attach to an existing view session re-installs the same hook idempotently.

## Test plan

- [x] `npx vitest run src/orchestrator.test.ts` — new assertions cover hook line emission, ordering (after `new-session`, before `exec attach`), and per-window scoping
- [x] Full op-obsidian suite green (`npx vitest run` — 772 tests)
- [x] `node scripts/bump-version.mjs patch` — lockstep bump to `0.61.5` + production build OK
- [x] `node scripts/dev-sync.mjs` + smoke test in OP-Test (plugin loaded, 39 commands registered, dev:console clean)
- [ ] Reviewer: walk through the bash-quoting of the hook string — `shSingleQuote` of a tmux command string with embedded single-quoted format predicate

## Adversarial review request

1. **Quoting correctness.** The bash line is `tmux set-hook -t '=view-OP-178' window-unlinked '<hookCmd>'` where `<hookCmd>` is `shSingleQuote("if-shell -F '#{==:#{hook_window_name},OP-178}' 'kill-session -t =view-OP-178'")`. Confirm the `'\''` escape sequence survives both bash dequoting and tmux's command-string parser, and that tmux receives exactly one argument with the inner single quotes still grouping the format predicate and the kill-session subcommand.
2. **Sibling-pane isolation.** When two agents (`OP-A`, `OP-B`) share an `op-agents-1` session and `OP-A` exits, the `window-unlinked` hook fires on every session that has it set — including `view-OP-B`. The format predicate `#{==:#{hook_window_name},OP-B}` should evaluate false in that fire and leave `view-OP-B` alone. Trace whether that is in fact how `#{hook_window_name}` resolves on the *receiving* session (vs. always reflecting the originating session).
3. **Re-attach idempotence.** When a user closes the iTerm pane and re-launches the same issue, `new-session … 2>/dev/null || true` no-ops because the view session already exists; `set-hook` re-runs and overwrites the previous hook. Confirm this is safe (no append-mode side-effect, no stale hook left behind from a prior issue id reusing the same session — though same-id reuse is the only path here).
4. **Race: agent dies before hook is installed.** The bash script runs `new-session`, then `set-hook`, then `printf … exec tmux attach`. If the agent's window is already dead when set-hook runs (e.g. agent crashed before pane attached), what happens? Does `set-hook` succeed against a session whose only window is gone? Does the hook fire retroactively on the next event?
5. **Manual `tmux unlink-window` from the user.** Per the issue body "if the user manually unlinks the agent's window via tmux keybindings, the hook still fires — same end state as a clean /quit, which is correct." Verify.
6. **Cleanup symmetry with `agentSessionCleanup.ts`.** That module also tears down view sessions. Does the new hook race or double-kill against it?

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.61.5",
+  "version": "0.61.6",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.61.4",
+  "version": "0.61.5",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.61.5",
+  "version": "0.61.6",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.61.4",
+  "version": "0.61.5",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/orchestrator.test.ts
+++ b/plugins/op-obsidian/src/orchestrator.test.ts
@@ -163,4 +163,35 @@ describe("buildViewScript", () => {
       `exec '/opt/homebrew/bin/tmux' attach -t 'view-OP-172' \\; select-window -t 'view-OP-172':'OP-172'`,
     );
   });
+
+  // OP-178: the per-pane view session inherits the parent op-agents-N
+  // session's window list, so the agent's `/quit` would otherwise leave the
+  // iTerm pane attached and silently switched to a sibling agent's window.
+  // The hook kills the view session when the agent's specific window is
+  // unlinked, so the iTerm pane closes cleanly.
+  it("installs a window-unlinked hook on the view session that kills it when the agent's window dies", () => {
+    const out = buildViewScript(baseArgs);
+    expect(out).toContain(
+      `'/opt/homebrew/bin/tmux' set-hook -t '=view-OP-172' window-unlinked 'if-shell -F '\\''#{==:#{hook_window_name},OP-172}'\\'' '\\''kill-session -t =view-OP-172'\\'''`,
+    );
+  });
+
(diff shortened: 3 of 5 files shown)


earchibald:
@copilot please run the adversarial review using the pressure-test prompts in the PR description. Particular focus:

1. **Bash + tmux quoting trace.** Walk through the line `'/opt/homebrew/bin/tmux' set-hook -t '=view-OP-178' window-unlinked 'if-shell -F '\''#{==:#{hook_window_name},OP-178}'\'' '\''kill-session -t =view-OP-178'\'''`. After bash dequoting, what does tmux's command parser receive as the value of the hook? Confirm the inner single-quote groups are preserved so the format predicate and the kill-session subcommand are parsed as separate tmux args, not concatenated into one malformed token.

2. **`#{hook_window_name}` resolution scope.** When `op-agents-1` has windows `OP-A`, `OP-B`, `OP-C`, each pane has its own grouped session (`view-OP-A`, `view-OP-B`, `view-OP-C`) with the hook installed. When `OP-A` is unlinked, the `window-unlinked` event fires. Does it fire only on `view-OP-A` (the session whose attached client was on that window), or does tmux fan it out to every grouped session that shares the window list (`view-OP-B`, `view-OP-C` too)? If the latter, the predicate `#{==:#{hook_window_name},OP-B}` must evaluate to false on the `view-OP-B` fire — confirm that's how tmux resolves `#{hook_window_name}` in fanned-out fires.

3. **Race: agent dies before the hook is installed.** The bash script order is `new-session` → `set-hook` → `printf` → `exec attach`. If the agent's window dies *between* `new-session` and `set-hook` (e.g. crash on first launch), `set-hook` runs against a session that exists but whose only window is gone. Does `set-hook` succeed? Will the hook fire on a future `window-unlinked` (none coming, since the only window is already gone), leaving the pane attached to an empty session forever? Suggest a defensive fallback if so.

4. **Re-attach idempotence — and the user-renames-issue case.** If the issue file is renamed (id stays `OP-178`), `view-OP-178` is reused, `set-hook` overwrites — fine. But if a user manually edited the in-progress issue's id to a different value, `tmuxWindow` would change while `groupSessName` would too — both substitute fresh, so the hook is reinstalled correctly. Verify this isn't a footgun.

5. **Cleanup symmetry with `agentSessionCleanup.ts`.** Does that module already kill `view-<id>` sessions? If yes, the new hook racing it produces "no server running" errors which the existing code may or may not swallow. Trace the interaction.

6. **Backwards compat.** Existing in-flight panes attached to view sessions created by an older `main.js` version will not have the hook installed. Restart-detach on a re-launch will install the hook cleanly. Confirm there's no path that would leave an old-version session permanently un-hooked.

PR #220, opened 2026-04-26
OP-182: document non-iTerm terminals as second-class and untested

### LINKED ISSUE #218: document non-iTerm terminal clients as second-class and untested
Tracked as op issue **OP-182**.

earchibald:
## Summary
- README "Terminal" section now spells out three tiers: **iTerm** (primary daily-driver), **Terminal.app** (second-class fallback), **other terminals** (Kitty/Alacritty/Ghostty/WezTerm — not supported, op routes through one of the two regardless).
- Settings tab "Terminal app" tooltip mirrors the same tiering so users see it without leaving the dropdown.
- Docs and one tooltip string only — no behavior change.

Tracks op issue OP-182. Closes #218.

## Test plan
- [x] `node scripts/bump-version.mjs patch` → 0.61.7, build green.
- [x] `node scripts/dev-sync.mjs` → reload op-obsidian in OP-Test.
- [x] Probe rendered Settings tab via `app.setting.openTabById("op-obsidian")` — Terminal-app row found, description starts "iTerm is the primary, daily-driver terminal —…".
- [x] `obsidian vault=OP-Test dev:console` → no errors after reload + settings render.

### DIFF
--- a/plugins/op-obsidian/README.md
+++ b/plugins/op-obsidian/README.md
@@ -96,9 +96,17 @@ Per-project repo paths (slug → absolute path). Overridden by a `repo_path:` in
 
 ### Terminal
 
-- **Terminal app** — `Terminal` (plain tmux) or `iTerm` (`tmux -CC` control mode). All agents share one tmux session (`op-agents`), one window per issue, so agents survive the terminal closing (reattach with `tmux attach -t op-agents`).
+**Supported terminals.** macOS only. Three tiers:
+
+- **iTerm — primary, daily-driver.** All agent-launch UX (tmux `-CC` control mode, the layout orchestrator, non-activating background launch via `open -ga`, recommended-prefs nudges, AppleScript automation) is built and tested against iTerm. Pick this unless you have a reason not to.
+- **Terminal.app — second-class fallback.** Works via plain tmux: each launch writes a launch script and `open -a Terminal`s it. No orchestrator, no `-CC` integration (each new Terminal window opens its own tmux client and mirrors the shared session), no background launch (Terminal has no equivalent non-activating cold-start), no pane layouts. Functional, but a degraded experience — only use it if iTerm isn't an option.
+- **Other terminals (Kitty, Alacritty, Ghostty, WezTerm, …) — not supported.** The dropdown is hardcoded to `Terminal | iTerm`; nothing else has been tested. If one of these is your default terminal, op still routes via `open -a Terminal` or iTerm — your preferred terminal is bypassed entirely. Adding support is not on the roadmap.
+
+Settings:
+
+- **Terminal app** — `Terminal` or `iTerm` (see tiers above). All agents share one tmux session (`op-agents`), one window per issue, so agents survive the terminal closing (reattach with `tmux attach -t op-agents`).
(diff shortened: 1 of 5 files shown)


earchibald:
@copilot please run an adversarial review on this docs-only PR. Particular focus:

1. **Tier accuracy vs. code reality.** Cross-check the README and Settings tooltip claims against the actual launch pipeline:
   - `plugins/op-obsidian/src/terminalLaunch.ts` — does the iTerm path *actually* gate orchestrator (`args.orchestrator?.settings.orchestrator.enabled`), `backgroundLaunch` (`args.terminalApp === "iTerm"`), and `iTermPlacement` (`new-tab` / `new-window`) all on `terminalApp === "iTerm"` only? If any of these are reachable from the Terminal.app branch (the `open -a Terminal` line at the bottom), the README's "no orchestrator / no `-CC` / no background launch / no pane layouts for Terminal.app" claim is wrong.
   - `plugins/op-obsidian/src/settingsPure.ts` `mergeSettings` — does it actually reject anything other than `"Terminal" | "iTerm"`? If a user pokes `data.json` to set `terminal: "kitty"`, does the load path really fall back, or does it leak through? The README says other terminals are "not supported" — that's only true if the merge clamps.
   - `renderDailyTerminal` in `settings.ts` — only emits `Terminal | iTerm`? Confirm the dropdown can't surface a third option via overlays or i18n.

2. **Settings tooltip rendering.** The new tooltip description is one long sentence with markdown backticks. Obsidian's `Setting.setDesc` renders as plain text by default — does the user see ` ` `tmux -CC` ` ` literally with backticks, or does Obsidian render it as code? If literal, decide whether that's acceptable or whether the description should be plain prose. Also check there's no length cap that truncates the tooltip in the actual rendered tab.

3. **Terminology drift inside the same file.** README §"### Terminal" now says "Other terminals … are not supported." But §"### iTerm layout orchestrator" still says "macOS + iTerm only" and the `setName("iTerm window placement")` description still reads "Where to open the agent when Terminal app is iTerm." Are these consistent with the new tier framing, or do they contradict it (e.g. by implying Terminal.app is a peer choice)?

4. **First-line-only stripped to 200 chars.** The smoke-test snippet truncates `desc.textContent` to 200 chars. Is the sentence I wrote actually the first 200 chars of the rendered text? (i.e. did Obsidian add a leading space or icon character that would shift the truncation and mask a typo at the start?)

5. **CLAUDE.md "after any change to plugins/op-obsidian/" loop.** Bump-version was run, dev-sync ran, smoke probe found the new copy. Is anything else listed in CLAUDE.md §1–§5 missing here? In particular, are there any required smoke probes I should have run for the daily-group rows after touching `renderDailyTerminal`?

6. **Backwards compat for the docs claim.** README says "If one of these is your default terminal, op still routes via `open -a Terminal` or iTerm — your preferred terminal is bypassed entirely." Is that *literally* true? Trace: when `terminal: "Terminal"` is set, the launch pipeline calls `/usr/bin/open -a Terminal <script>`. macOS's `open -a Terminal` always launches *Terminal.app* regardless of the user's default terminal handler — but is there any path (LSHandlerForContentType for `.command`, a plist override, etc.) where this would route through a different terminal app the user has set as default for `.command` files? If yes, the README is overstating. If no, confirm.

PR #256, opened 2026-04-26
OP-194: generic {{var}} renderer + pluginVarRegistry + WorkflowDiagnostic

earchibald:
## Summary

Foundation for the workflow-modules engine (parent: OP-184; grandparent: OP-181). Pure-fn only, no vault IO — the loader is 1b, composition is 1d.

- **`workflowDiagnostic.ts`** — declares the full `WorkflowDiagnostic` union from OP-181 plan §"Unified diagnostic record" (`bad-model | missing-var | unknown-module | schema-mismatch | import-collision | intra-scope-collision | malformed-frontmatter`). 1a only emits `missing-var`; 1b/1c add emit sites without re-editing the union shape.
- **`pluginVarRegistry.ts`** — single source of truth for the 16 always-on plugin vars (`id, title, project, status, priority, parent, repo_path, vault_path, vault_name, branch, pr_url, github_issue, today, agent, model, mode`). Each entry: `{ name, description, example, compute(ctx) }`. `PARENT_NONE_SENTINEL = "(none — this is a top-level issue)"` lives here so every consumer references one source. `buildRenderContext({ entry, profile, launch })` flattens `IssueEntry + AgentProfile + LaunchContext` into a typed `RenderContext`. No I/O, no clock reads.
- **`renderTemplate.ts`** — generic `{{var}}` pure renderer. Token shape `{{ \s* [a-zA-Z_][a-zA-Z0-9_]* \s* }}`, registry-driven resolution. Unknown name or undefined compute → leaves the token verbatim and emits `WorkflowDiagnostic { code: "missing-var", varName }`. Verbatim fallback is the explicit safety contract: never silently corrupts a prompt with empty strings, never throws on a typo.
- **`agentProfiles.ts:renderSkillTrigger`** — refactored to delegate to `renderTemplate`. Public signature preserved; output byte-identical for every shipped trigger string (asserted in tests).

Bumps `op-obsidian` to `0.62.0` (minor — additive infrastructure; no user-visible behavior change yet).

## Test plan

- [x] `npm test` in `plugins/op-obsidian/` — 836 tests pass across 60 files. The one pre-existing `iTermPrefs.test.ts` infra failure reproduces on the unmodified branch tip; unrelated.
- [x] Targeted run of new + adjacent files: 84 / 84 passing (`pluginVarRegistry.test.ts`, `renderTemplate.test.ts`, `agentProfiles.test.ts`).
- [x] `node scripts/bump-version.mjs minor` — manifest, package.json, plugin.json all at 0.62.0; build green.
- [x] `node scripts/dev-sync.mjs` — copied into OP-Test; plugin reloaded clean.
- [x] OP-Test smoke probe — plugin loaded at v0.62.0, all 39 commands registered, no console errors.

### Test coverage highlights

- Every required always-on var has an entry (acceptance criterion).
- `parent` returns the literal id when present, `PARENT_NONE_SENTINEL` when null. The sentinel is sourced from the registry export — re-typing the prose at the assertion site would defeat the test.
- `renderTemplate`: single / multi / repeated tokens, whitespace-tolerant braces, mixed resolved + missing in one pass, idempotency (rendering the renderer's own output is a fixed point for fully resolved strings), `{{vars.<name>}}` shape ignored (left for 1b/1c), parent sentinel propagation through the renderer.
- `renderSkillTrigger` byte-equivalence asserted for `claude`, `gemini`, `copilot` profiles plus a synthetic `{{id}} {{id}} {{id}}` profile (matches legacy `/g` semantics).

## Out of scope (handled by sibling issues)

- Module file schema, frontmatter parser, loader → OP-195 (1b).
- Workflow file schema, `modelRegistry`, `extends:` inheritance, legacy fallback ladder → OP-196 (1c).
- IO loader (`loadModuleSources`), composer split (`composeWorkflow`), `RelaySession`, `launchHeadlessSubtask` rename → OP-197 (1d).
- The `{{vars.<name>}}` user-var namespace and precedence resolution → 1b/1d.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.61.8",
+  "version": "0.62.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.61.8",
+  "version": "0.62.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/agentProfiles.ts
+++ b/plugins/op-obsidian/src/agentProfiles.ts
@@ -1,3 +1,5 @@
+import { renderTemplate } from "./renderTemplate";
+
 export type AgentId = "claude" | "gemini" | "copilot";
 
 export const AGENT_IDS: AgentId[] = ["claude", "gemini", "copilot"];
@@ -244,7 +246,13 @@ export function mergeProfile(id: AgentId, overlay?: ProfileOverlay): AgentProfil
 }
 
 export function renderSkillTrigger(profile: AgentProfile, issueId: string): string {
-  return profile.skillTrigger.replace(/\{\{id\}\}/g, issueId);
+  // Delegates to the generic `renderTemplate` so every prompt-shaping
+  // surface goes through one code path with one diagnostic contract. The
+  // shipped trigger strings only reference `{{id}}` today, so a partial
+  // context with just `{ id, agent }` is sufficient. Diagnostics are
+  // discarded here; 1d's composer will plumb them through to the dry-run /
+  // settings surfaces.
+  return renderTemplate(profile.skillTrigger, { id: issueId, agent: profile.id }).text;
 }
 
 export function launchFlagsFor(profile: AgentProfile, mode: AgentLaunchMode): string[] {
--- a/plugins/op-obsidian/src/pluginVarRegistry.test.ts
+++ b/plugins/op-obsidian/src/pluginVarRegistry.test.ts
@@ -182,6 +182,18 @@ describe("PLUGIN_VAR_REGISTRY", () => {
       expect(PLUGIN_VAR_REGISTRY.parent.compute(minimal)).toBe(PARENT_NONE_SENTINEL);
       expect(PLUGIN_VAR_REGISTRY.parent.compute({ ...minimal, parent: "OP-2" })).toBe("OP-2");
     });
+
+    it("{{parent}} returns undefined (not the sentinel) when parent slot is missing from a Partial ctx", () => {
+      // parent is required (string | null) in a full RenderContext, but in a
+      // Partial<RenderContext> the slot may simply be absent (undefined).
+      // undefined means "caller didn't provide the parent field at all" — the
+      // renderer treats that as missing-var, not as a confirmed top-level issue.
+      // null is the explicit signal that the issue has no parent; only that
+      // triggers the sentinel.
+      const noParentSlot: Partial<typeof minimal> = { ...minimal };
+      delete (noParentSlot as Partial<typeof noParentSlot>).parent;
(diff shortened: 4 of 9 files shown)


earchibald:
@copilot — please adversarially review this PR. Concrete pressure-test prompts:

**Renderer correctness**
1. **Idempotency**: `renderTemplate(renderTemplate(text, ctx).text, ctx)` should equal `renderTemplate(text, ctx)` for any input. Construct a counterexample if you can — e.g., a registry value that itself contains a `{{var}}`-shaped substring (`title: "Use {{id}} in prose"`), or a Unicode/escape edge case the regex mishandles.
2. **Non-greedy regex / pathological inputs**: `{{ {{id}} }}` — what happens? `{{ id`, `id }}`, `{{}}`, `{{ }}` — verify no infinite loop, no exception, no spurious diagnostic.
3. **Whitespace edges**: `{{\nid\n}}` — the regex uses `\s*` which matches `\n`. Is that desired or should it reject newlines inside braces?
4. **Token-name collisions**: registry has `pr_url`, `github_issue`. What about `{{prurl}}`, `{{github-issue}}`, `{{vars.id}}`? The PR claims `vars.<name>` is left for 1b/1c; verify that actually holds (no false-positive substitution, no false-negative diagnostic).

**Registry contract**
5. **Sentinel duplication**: grep the diff for the literal string `(none — this is a top-level issue)`. It should appear *only* in `pluginVarRegistry.ts` as the `PARENT_NONE_SENTINEL` constant. Any other occurrence (test fixture, doc string, comment) is duplication the PR claims to prevent.
6. **`parent: string | null` boundary**: `compute({ parent: undefined })` (Partial ctx, parent slot missing) vs. `compute({ parent: null })` vs. `compute({ parent: "OP-1" })` — the docstring promises sentinel-on-null and the literal id otherwise. Verify the `undefined` case (the type allows it via `Partial<RenderContext>`) returns `undefined` (which would then emit a missing-var diagnostic in the renderer) — is that the intended UX, or should the registry treat `undefined` and `null` the same way?
7. **`buildRenderContext` mapping**: `IssueEntry.pr → pr_url`, `IssueEntry.githubIssue → github_issue`, `AgentProfile.id → agent`. Any other field that *should* flow through but doesn't? E.g., `IssueEntry.assignee`, `IssueEntry.priority` (priority is mapped — verify the test covers a missing-priority case).
8. **Type drift on `WorkflowDiagnostic`**: the union declares 7 codes but only 1 is emitted. If a future PR adds a code to the union but forgets to handle it in some consumer's `switch`, TypeScript should catch it. Is the type set up for exhaustiveness checks?

**Backwards compatibility**
9. **`renderSkillTrigger` byte-equivalence**: the legacy `replace(/\{\{id\}\}/g, issueId)` and the new `renderTemplate(...).text` must produce identical output. Tests assert this for the three base profiles. Construct an overlay or input where they could diverge — e.g., an `issueId` containing `$&` or other regex-replacement special chars (the `String.prototype.replace` callback form should be safe, but worth a sanity probe).
10. **Circular imports**: `agentProfiles → renderTemplate → pluginVarRegistry → agentProfiles` (type-only). The PR notes `import type` strips at compile, but verify the built `main.js` doesn't have a runtime cycle (the bundler may inline or hoist in unexpected ways).

**Test rigor**
11. The PR's tests are descriptive but: are there assertions that would *only* fail on a buggy implementation? Or do any tests assert tautologies (e.g., "compute(ctx) returns ctx.x" where compute is just `(ctx) => ctx.x`)? Flag any test where the assertion mirrors the implementation rather than the contract.
12. `it("does no I/O — pure assembly given the inputs")` in `pluginVarRegistry.test.ts` — this is two equality calls. Is that actually testing purity? Suggest a stronger formulation if possible.

Take your time — this is foundation code for OP-181 and 1b/1c/1d build on top of it. Better to surface a contract bug here than have it ripple into four downstream PRs.

PR #257, opened 2026-04-26
OP-195: workflow-module schema + loader + typed VarDecl parser

earchibald:
## Summary

- Pure-fn shape: `parseVarDecl` / `parseVarDecls` / `parseModule` / `validateIntraScopeCollisions` in `workflowModulePure.ts` — no Obsidian imports, no IO. Three `VarDecl` shapes (`bare`, `default`, `object`) with table-driven coverage of every edge from the issue scope.
- Thin IO seam in `workflowModule.ts`: `loadModules(app, opts?)` walks `Projects/_op-modules/` and `Projects/<slug>/MODULES/`, applies per-project shadowing, optional project filter, runs intra-scope collision validation.
- Diagnostics: `malformed-frontmatter` for invalid declarations (wrong type, id-vs-filename mismatch, non-string fields, YAML-coerced Date, duplicate var name within one module); `intra-scope-collision` after shadowing.
- Schema reference: `docs/specs/workflow-module-schema.md` — single source for users authoring modules; covers shadowing surprise, YAML auto-coercion gotcha, diagnostic table.
- Bumps op-obsidian `0.62.0 → 0.63.0` (minor — additive infra; no user-visible behavior change yet).

No composition. No `extends:`. No `modelRegistry`. No legacy `WORKFLOW.md` fallback. All deferred to 1c/1d (OP-196/OP-197).

## Test plan

- [x] 51 pure tests in `workflowModulePure.test.ts` covering: bare names, name=VALUE shorthand, name= (empty default), name=VALUE with `=`-in-value, quoted values, YAML-coerced types stay strings, Date rejection, object form with description, missing names, duplicate-name-within-one-module, parseModule happy path + every required-field rejection, vars-level diagnostics don't disqualify the module, intra-scope-collision (2-way, 3-way, no-cross-scope, one-per-var).
- [x] 12 IO tests in `workflowModule.test.ts` covering: path bucketing, ignoring out-of-band files, per-project shadowing of same-id globals, project filter behavior, malformed-frontmatter on bad files, collision check runs against post-shadowing set.
- [x] Full suite: 916 / 916 passing (the 1 failing suite is `iTermPrefs.test.ts` — pre-existing infra issue with `obsidian` package resolution in vitest, also failing on unmodified main; unrelated to this PR; documented in OP-194).

### DIFF
--- a/docs/specs/workflow-module-schema.md
+++ b/docs/specs/workflow-module-schema.md
@@ -65,10 +65,12 @@ generic {{var}} template engine from OP-194.
 - **`project`** (string) — restricts a global module to a specific project
   slug. Per-project modules already live under their slug, so they don't need
   this field; if set, the value should match the enclosing slug or the module
-  is filtered out at load time.
+  is filtered out at load time. An empty or whitespace-only value (`project:
+  ""`) is silently treated as absent — the field is ignored and the module
+  applies to all projects.
 - **`agent`** (string) — restricts the module to a specific agent id
   (`claude`, `gemini`, `copilot`, custom). Loader doesn't filter on this
-  today; consumers may apply it.
+  today; consumers may apply it. Same empty-string rule as `project`.
 - **`order`** (integer, default `0`) — sort key within a scope.
 - **`vars`** (list of `VarDecl`) — see below.
 
@@ -101,6 +103,12 @@ The first `=` splits name from default value. The empty-default form
 caller must supply a value"; empty-default means "the default is the empty
 string".
 
+**Value whitespace is preserved verbatim.** `bar=  hi  ` yields a default
+of `"  hi  "` (two leading spaces, two trailing). If your intent is "empty
+default", write `bar=` (no trailing space). Spaces inside the name
+segment (before `=`) are trimmed, but spaces in the value segment
+(after `=`) are not.
+
 #### YAML auto-coercion gotcha
 
 YAML coerces some bare scalars to non-string types — most commonly ISO dates
@@ -125,6 +133,18 @@ vars:
 `name:` is required and must be a non-empty string. `default:` and
 `description:` are optional and must be strings. Unknown keys are tolerated
 (forward-compat for future fields like `enum:` or `required:`).
+
+**Watch out for typos.** `defualt:` vs `default:` won't emit an error —
+unknown keys are silently accepted. If a default value appears to be
+ignored, double-check the key spelling.
+
+### Var name identifiers
+
+Variable names are not validated against an identifier regex at parse time.
+However, the OP-194 renderer uses `[a-zA-Z_][a-zA-Z0-9_]*` to match
+`{{var}}` placeholders — names containing hyphens or starting with a digit
+will parse and load successfully but will **not** be expanded by the renderer.
+Use names that match `[a-zA-Z_][a-zA-Z0-9_]*` for full compatibility.
 
 ### Duplicate names
 
--- a/docs/specs/workflow-module-schema.md
+++ b/docs/specs/workflow-module-schema.md
@@ -1,0 +1,185 @@
+# Workflow Module Schema
+
+Status: shipped in OP-195 (1b of OP-184). Composition (1d/OP-197) consumes
+loaded modules; this doc covers the file format and load behavior only.
+
+## What a module is
+
+A workflow module is a markdown file whose frontmatter declares a partition of
+the composed workflow — typically a step (or set of steps) that should be
+spliced into a larger workflow at a named scope. Composition stitches modules
+together; **this doc covers the file format only**, not how composition works.
+
+## Where modules live
+
+| Location | Path | Applies to |
+| :--- | :--- | :--- |
+| Global | `Projects/_op-modules/<id>.md` | Every project. |
+| Per-project | `Projects/<slug>/MODULES/<id>.md` | Only project `<slug>`. |
+
+`<id>` is the module id. **The filename basename must match the `id:`
+frontmatter field exactly.** A file named `review-and-merge.md` must declare
+`id: review-and-merge`. Mismatches emit `malformed-frontmatter` diagnostics
+and the module is dropped.
+
+Files outside these two layouts are ignored — including subdirectories. A
+file at `Projects/_op-modules/sub/foo.md` is **not** loaded.
+
+## Frontmatter contract
+
+```yaml
+---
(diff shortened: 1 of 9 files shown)


earchibald:
@copilot — please adversarially review this PR. Concrete pressure-test prompts:

**VarDecl parser correctness**

1. **First-`=` split semantics**: `parseVarDecl("name=foo=bar")` should yield `value: "foo=bar"` (only first `=` splits). Construct an input where this fails — e.g., a name containing literal `=` (impossible by design, but verify the docstring matches behavior). Also: `parseVarDecl("=")` (no name, no value, just `=`) — what does it emit? The code path is empty-name → `malformed-frontmatter`, but verify there's no off-by-one that lets `name=""` slip through.

2. **Bare-vs-empty-default distinction**: `{ kind: "bare", name: "x" }` and `{ kind: "default", name: "x", value: "" }` are intentionally different. A bare means "caller must supply"; an empty-default means "default is empty string". Construct a test where the composer (1d) would behave differently for the two — and verify nothing in the parser silently collapses them. Also check: what does `parseVarDecls(["x", "x="])` return? Per the duplicate rule, the first wins (so the bare survives, the empty-default gets dropped) — is that the right precedence, or should the more-specific declaration win?

3. **Whitespace handling asymmetry**: `parseVarDecl("  foo  ")` trims (yields `name: "foo"`). `parseVarDecl("name=  hi  ")` does **not** trim (yields `value: "  hi  "`). Defensible? The schema doc claims "value preserves leading/trailing whitespace" but a user typing `bar=   ` to mean "default is empty" gets a default of `"   "` instead. Surface this UX gotcha or argue why it's correct.

4. **YAML auto-coercion edges**: the canonical test is `due=2026-04-25` (string, splits on `=`). What about:
   - `parseVarDecl("2026-04-25")` (bare ISO date, no `=`) — YAML would have coerced this to `Date` before the parser sees it. The string form returns `{ kind: "bare", name: "2026-04-25" }` — is "2026-04-25" a valid var name? The parser doesn't validate the name shape against an identifier regex; should it? OP-194's renderer regex is `[a-zA-Z_][a-zA-Z0-9_]*` — names with hyphens or starting with digits won't render even if they parse. Consider adding name validation here vs. deferring to the renderer.
   - `parseVarDecl("name=true")`, `parseVarDecl("name=42")` — both succeed as `value: "true"` / `value: "42"`. Confirm the YAML layer actually delivers these as strings (vs. auto-coercing to bool/number before the parser sees them). This is testable only at the YAML boundary, not the parser — flag if YAML *does* coerce.

5. **Object-form unknown-key tolerance**: `parseVarDecl({ name: "x", futureField: 42 })` is silently accepted. Argue why this is forward-compat (1c/1d may add `enum:` / `required:` / `pattern:`) versus why it could mask typos (`defualt:` vs `default:` won't error). Should there be a "warn-but-accept" mode that emits an `info`-severity diagnostic for unknown keys, so typos show up in the dry-run?

**Module parser correctness**

6. **Id-vs-filename mismatch**: `parseModule({ id: "sample", frontmatter: { id: "different", ... }})` rejects with `malformed-frontmatter`. What if `id:` is missing entirely from frontmatter (vs. set to a wrong value)? The code path for missing `id` should still surface a clear error. Trace it.

7. **Hard-fail vs. soft-fail boundary**: `parseModule` returns `module: null` only when a "hard fail" (missing required field, type mismatch, id mismatch) hits. Vars-level diagnostics are "soft" — bad var entries are dropped, the module survives. Is the slicing logic in `parseModule` (`diagnostics.slice(0, diagnostics.length - varsResult.diagnostics.length)`) correct, or could it miscount when a hard-fail diagnostic and a vars diagnostic are emitted in the same parse? Construct an input where this could matter.

8. **Optional-field empty-string handling**: `frontmatter.project = ""` (empty string, present but empty). The code path treats this as `malformed-frontmatter` (empty string is non-empty fail). But frontmatter authors might intentionally set `project: ""` to mean "no project". Should this be tolerated as `project: undefined`? Argue both sides.

**IO loader correctness**

9. **Path-bucketing depth**: globals must live exactly one level deep below `Projects/_op-modules/`. `Projects/_op-modules/sub/foo.md` is silently dropped. Is silent drop correct, or should the loader emit an `info`-severity diagnostic ("workflow modules in subdirectories are not loaded")? Same question for `Projects/<slug>/MODULES/sub/foo.md`.

10. **Per-project file with `project:` field mismatch**: `Projects/obsidian-projects/MODULES/foo.md` declares `project: different-project`. The loader's filter (`!m.project || m.project === opts.project`) would drop this when `opts.project = "obsidian-projects"`. Should the loader instead emit a `malformed-frontmatter` for "per-project module's `project:` field doesn't match its enclosing slug"? Today it's silently filtered. Argue both sides.

11. **Shadowing UX**: the schema doc warns about silent shadowing. Is there a CLI / dry-run / settings surface today that surfaces "module X shadows global Y"? If not, file a follow-up — users can lose hours wondering why a global module went quiet.

**Diagnostics shape**

12. **`extra` field consistency**: `malformed-frontmatter` from `parseModule` carries `extra.path`, `extra.field`, `extra.expected`, `extra.actual`. `malformed-frontmatter` from `parseVarDecls` (when wrapped by `parseModule`) carries `extra.path`. Consumers (1d's settings UI) must know which fields are guaranteed. Is the contract documented? If not, add a TSDoc note on `WorkflowDiagnostic.extra` listing the keys per code.

13. **`intra-scope-collision` extra fields**: carries `extra.scope` and `extra.moduleIds`. The diagnostic message also names the modules. Risk: the message and the `extra` may drift if the message-formatting changes — a downstream consumer parsing the message string instead of the structured `extra` would silently break. Verify there are no consumers parsing the message string.

**Test rigor**

14. The pure tests are table-driven and exhaustive, but: is there a property-based test (fuzz) that would catch a regression the table doesn't? Suggest one — e.g., "for any `parseVarDecl` input, the result satisfies `decl !== null XOR diagnostics.length > 0`" (mutual exclusion of success and diagnostic). Verify this invariant holds in the current code.

15. The IO tests use a hand-rolled fake `App`. Does any test exercise the `getMarkdownFiles()`-returns-many-files path with both globals and per-project files mixed in (verifies the bucketing loop's order-independence)? Check.

Take your time — this is foundation code for OP-184, and 1d (OP-197) builds the composer on top. Better to surface a contract bug here than have it ripple into the composer PR.

PR #258, opened 2026-04-26
OP-196: workflow file schema + modelRegistry + extends inheritance + legacy fallback

earchibald:
## Summary

Workflow-file schema, IO loader, and per-agent model registry. Third grandchild of OP-184 (Child 1 umbrella) after OP-194 (1a, generic renderer) and OP-195 (1b, module schema); composition (1d / OP-197) is the first consumer.

- `workflowFilePure.ts` — pure types + parsers (`parseAgentSpec`, `parseModelSpec`, `parseWorkflowFile`), legacy classifier + synthesizer, `validateWorkflowModels`. No Obsidian imports.
- `workflowFile.ts` — IO seam: reads `Projects/<slug>/WORKFLOW.md`, applies the legacy fallback ladder, recursively loads the parent at most one level for `extends:`, runs model validation.
- `modelRegistry.ts` — per-agent `{ aliases, versioned }` records for `claude`, `gemini`, `copilot`. Single source of truth for model validation. `BadModelSpec` payload doubles as the `bad-model` `WorkflowDiagnostic.extra` shape.
- Six legacy-workflow fixtures + per-shape unit tests (`__fixtures__/legacy-workflow/`).
- `docs/specs/workflow-file-schema.md` — repo-tracked schema reference.

132 net new tests (105 pure + 18 IO + 9 fixture); full suite 1056/1056; the one failing suite (`iTermPrefs.test.ts`) is the same pre-existing `obsidian` package resolution issue OP-194/195 documented, unrelated.

Bumps op-obsidian `0.63.1 → 0.64.0` (minor — additive infra).

## Test plan

- [x] `npm test` in `plugins/op-obsidian/` — 1056/1056 pass.
- [x] `node scripts/bump-version.mjs minor` — version moves in lockstep across `manifest.json`, `package.json`, `plugin.json`; `main.js` rebuilt and fresher than `manifest.json`.
- [x] `node scripts/dev-sync.mjs` — main.js + manifest.json copied to OP-Test, plugin reloaded.
- [x] OP-Test smoke: `obsidian eval` confirms `op-obsidian@0.64.0` instance with 39 commands registered, no console errors. New types are cold infrastructure (no callsite wires them up yet) so runtime behaviour is unchanged.

## Adversarial review prompts (Copilot)

Please pressure-test:

1. **Parser shape edges**. Can `parseAgentSpec` / `parseModelSpec` produce a `value` that is non-null **and** carries diagnostics? When? Should it? (Look at `kind: "perAgent"` with one valid + one bad entry.) Any input that passes shape validation but is semantically wrong (e.g. `default_agent: ["", "claude"]` — does the silent drop of blank entries hide a typo)?
2. **`schema: 1` lock**. Today this loader hard-rejects any other `schema:` value. Is the diagnostic message specific enough that a future migration tool can be built around it? Does the wording survive being shown to an end-user in a Notice?
3. **Legacy ladder ambiguity**. Shape 5 (null frontmatter) vs. shape 2 (frontmatter without `type:`): for an empty fence `---\n---`, does `metadataCache` reliably return `null`, or does it sometimes return `{}`? Verify the IO disambiguation logic in `workflowFile.ts:loadAtPath` (the `fmCached === undefined ? null : fmCached` branch) — is there a Obsidian frontmatter shape that lands in the wrong shape bucket?
4. **`extends:` cycles**. The recursive call passes `allowExtends: false` to enforce one-level. What if the **child's** `extends:` points at itself (`Projects/foo/WORKFLOW.md` → `Projects/foo/WORKFLOW.md`)? The recursive call would re-read the same file with `allowExtends: false`; the child's `extends:` would then trigger the warning, but is there any way to enter an actual loop?
5. **Step merge by id**. `mergeWorkflows` substitutes a child step at the parent's slot when the id matches. What if a child wants to *remove* a parent step? (Today: can't — v1 limitation. Is the diagnostic / docs clear about this?) What if both the parent and the child have multiple steps with the same id (already a per-file error, but: does the merge still produce sensible output)?
6. **Model validation cross-product blast**. `validateWorkflowModels` for a workflow with `default_agent: [a, b, c]` and `default_model: [m1, m2, m3]` (`kind: "all"`) where `m1` is invalid for all three agents: should this emit 3 diagnostics or 1? Today it emits 3. Is that what an author wants?
7. **Per-agent map referencing absent agent**. We emit a warning when `default_model: { gemini: pro }` is set but `gemini` isn't in `default_agent`. But composition (1d) might *want* this if `default_agent: [claude]` and a step overrides to `agent: gemini` then inherits the `gemini: pro` from defaults. Is the warning-severity right, or should this be silent?
8. **Filename hygiene**. `synthesizeLegacyWorkflow` accepts arbitrary `path` / `project` strings — no validation. If a caller passes `path: "../escape"` does anything bad happen downstream? (Probably not — it's load-and-validate cold infra — but worth confirming.)
9. **`BadModelSpec` privacy**. `allowedAliases` / `allowedVersioned` arrays leak the entire registry to the diagnostic stream. If a future version has private/experimental models, is this surface tenable? (Probably want a "user-visible aliases" subset later.)
10. **Test fixtures vs. real Obsidian behaviour**. The fixture tests hand-build frontmatter objects to simulate `metadataCache` output. Does Obsidian's actual YAML parser produce exactly the shape we hand-build for shapes 2, 3, 5? Specifically: for an empty fence (`---\n---`), what does `metadataCache.getFileCache(file)?.frontmatter` return — `null`, `undefined`, or `{}`? The IO test should ideally lock the answer.

### DIFF
--- a/docs/specs/workflow-file-schema.md
+++ b/docs/specs/workflow-file-schema.md
@@ -1,0 +1,260 @@
+# Workflow File Schema
+
+Status: shipped in OP-196 (1c of OP-184). Composition (1d/OP-197) consumes the
+loaded workflow; this doc covers the file format, the `extends:` inheritance
+rule, the model registry, and the legacy-fallback ladder. Module composition
+itself lives in [`workflow-module-schema.md`](./workflow-module-schema.md).
+
+## What a workflow file is
+
+A workflow file is the per-project description of *what gets injected, by
+which agent, at which step*. It declares:
+
+- **Defaults**: the agent and model used when a step doesn't override them.
+- **Steps**: an ordered list of named stages — `kickoff`, `review`,
+  `finalize`, etc. Each step references a list of module ids (defined in
+  [workflow modules](./workflow-module-schema.md)) that supply its content
+  and variable declarations. A step may override the default agent or model.
+- **Inheritance**: an optional `extends:` pointer to a parent workflow file
+  (typically a global default at `Projects/_op-workflow.md`). Parent steps
+  apply unless the child repeats a `step:` id, in which case the child's
+  step replaces the parent's at that slot.
+
+## Where workflow files live
+
+| Location | Path | Applies to |
+| :--- | :--- | :--- |
+| Per-project | `Projects/<slug>/WORKFLOW.md` | Project `<slug>`. |
+| Global default | `Projects/_op-workflow.md` (by convention) | Any project that names it via `extends:`. |
+
+The loader doesn't auto-merge globals — a project must opt in by declaring
+`extends:` explicitly. This is intentional: an unnamed global is invisible to
+authors editing the per-project file.
+
+## Frontmatter contract
+
+```yaml
+---
+type: workflow                        # required, exact literal
+schema: 1                             # required integer; locks the contract
+project: obsidian-projects            # required string (project slug)
+default_agent: claude                 # list-or-scalar
+default_model: opus                   # list-or-scalar-or-keyed-map
+extends: Projects/_op-workflow.md     # optional, vault-relative; one level only
+steps:                                # required (in modern shape)
+  - step: kickoff                     # required, unique within the workflow
+    modules: [orient, identify-issue] # list of module ids; may be empty
+    agent: claude                     # optional override (list-or-scalar)
+    model: opus                       # optional override (list-or-scalar-or-keyed-map)
+  - step: review
+    modules: [review-and-merge]
+---
+
+# Markdown body is opaque commentary today. The synthetic legacy-kickoff step
+# is the only path that attaches body content to a workflow programmatically.
+```
+
+### `type` and `schema`
+
+`type` is the literal `workflow`. `schema` is an integer locking the format
+version. **This loader supports `schema: 1` only.** Files declaring
+`schema: 2` (or any other value) emit `schema-mismatch` and are dropped —
+future versions will be added with explicit migration paths.
+
+### `project`
+
+The project slug. Must be a non-empty string. If it disagrees with the
+file's enclosing folder name (`Projects/<slug>/`), the loader emits a
+warning-severity `malformed-frontmatter` diagnostic and uses the folder
+name. The warning is intentional: the folder is canonical, but a
+mistyped `project:` field is a useful breadcrumb when copying a workflow
+between projects.
+
+### `default_agent` (list-or-scalar)
+
+Accepts:
+
+- A scalar string: `default_agent: claude`. Equivalent to `[claude]`.
+- A list of strings: `default_agent: [claude, gemini]`. The launch context
+  picks one at runtime.
+
+Empty input, whitespace-only entries, or non-string entries emit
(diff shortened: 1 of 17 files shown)


earchibald:
@copilot please review against the 10 pressure-test prompts in the PR body. Particular focus areas where independent eyes would help:

- **Prompts 3 (legacy ladder ambiguity) and 10 (fixture vs. real Obsidian)**: the IO loader's `fmCached === undefined ? null : fmCached` disambiguation hinges on what Obsidian's `metadataCache.getFileCache(file)?.frontmatter` returns for various legacy shapes. If you can verify against the actual Obsidian metadataCache contract (or surface where it's documented), please call out any shape we'd misclassify.

- **Prompt 4 (`extends:` cycles)**: walk through what happens when a workflow `extends:` itself, or when shape 4 (wrong-type) is reached *via* `extends:` from a modern child.

- **Prompt 6 (cross-product blast)**: 3-agent × 3-model = 9 cross-products is the right count for `kind: "all"` validation, but if 5 of those fail an author sees 5 diagnostics that all point at the same `default_model:` line. Worth proposing a deduplication strategy.

- **Anywhere `parseModelSpec` semantics break user expectations**: the keyed-map shape silently drops bad entries while preserving good ones. Is that the right call, or should the whole field invalidate?

For anything you'd ship as a fix, push the commit; for anything that's a discussion call, comment with the trade-off and let the human decide. Skip the prose-only nits — go straight for behavioural pressure.

earchibald:
Reviewed @copilot's findings. Both shipped fixes are sound — pulled, re-tested (1058/1058 pass; +2 net new), rebuilt and re-synced to OP-Test, plugin reloads cleanly at 0.64.0.

**Shipped fixes — accepted as-is.**

- Self-extension guard: catches `extends:` pointing at the same file with a clear error-severity diagnostic (no longer the misleading "nested extends" warning that the absent guard would have surfaced via the recursive call). Test locks the message shape.
- `metadataCache` comment correction: matches the real `FrontMatterCache | undefined` contract — empty fence yields `undefined`, never `null`. The IO layer's `undefined → null` mapping for shape-1-vs-5 disambiguation is preserved and correct.

**Discussion items — defer per Copilot's recommendation.**

- **Prompt 1** (silent blank-entry drop): keep silent for now. The cost of warning on intentional trailing-comma YAML outweighs the rare leading-blank typo. Revisit if real-world WORKFLOW.md files surface false negatives.
- **Prompt 6** (cross-product diagnostic blast): defer the dedup to OP-197 (1d) caller-side. Adding an `agents: string[]` field to `BadModelSpec` is a minor additive change but carries diagnostic-stream interface churn before any consumer exists; better to let 1d's surface drive the shape with real composition context.
- **Prompt 7** (per-agent map warning when agent is absent from default_agent): keep warning as-is. 1d's step-agent resolution will be the right place to suppress when a step explicitly names that agent — the validator has no way to know that today without lookahead into composition.

No blockers. PR ready to merge from my side once you give the go-ahead.

Issue #222, opened 2026-04-26
Workflow module engine — schemas, loaders, renderer, composer (umbrella for 1a–1d)


earchibald:
Tracked as op issue **OP-184**.

## Scope

- [ ] - [ ] This is the foundation child issue for OP-181. It is itself an umbrella for grandchildren 1a–1d which will be spawned as separate child issues of THIS issue when work begins.
- [ ] - [ ] Grandchild 1a: generic {{var}} renderer + context builder + pluginVarRegistry.ts (single source of truth for always-on plugin vars including {{parent}} sentinel)
- [ ] - [ ] Grandchild 1b: Module file schema + loader (Projects/_op-modules/
- [ ] Projects/<slug>/MODULES/) with typed VarDecl parser ({ kind: bare | default | object }) + table-driven tests
- [ ] - [ ] Grandchild 1c: Workflow file schema + loader + modelRegistry.ts (per-agent alias + versioned-id source of truth) + list-or-scalar-or-keyed-map agent/model parsing + extends: workflow inheritance + legacy WORKFLOW.md fallback ladder (six shapes
- [ ] fixtures + per-shape tests)
- [ ] - [ ] Grandchild 1d: loadModuleSources (IO) + composeWorkflow (pure) split + RelaySession discriminated union + launchHeadless rename to launchHeadlessSubtask + end-to-end integration test golden snapshot
- [ ] - [ ] Sets parent: OP-181 on this issue. When this issue runs
- [ ] spawn 1a–1d as children of itself; the parent of 1a–1d is THIS issue
- [ ] not OP-181.
- [ ] - [ ] See OP-181's Plan section and docs/plans/OP-181-workflow-modules.md for full design.

PR #259, opened 2026-04-26
OP-197: composeWorkflow split (IO + pure) + RelaySession + launchHeadlessSubtask + integration test

### LINKED ISSUE #235: 1d: composeWorkflow split (IO + pure) + RelaySession + launchHeadlessSubtask + integration test
Tracked as op issue **OP-197**.

## Scope

- [ ] - [ ] Parent: OP-184 (Child 1 umbrella). This is the integration glue for 1a/1b/1c.
- [ ] - [ ] Two functions
- [ ] mirroring the codebase's *Pure.ts pattern: loadModuleSources(app): Promise<ModuleSource[]> (IO; reads vault TFiles for global + per-project module folders); composeWorkflow(sources
- [ ] workflow
- [ ] mode
- [ ] ctx): ComposedPrompt (pure
- [ ] sync; resolves precedence
- [ ] applies render pass
- [ ] returns { ordered chunks
- [ ] perVarSourceMap
- [ ] sizeChars
- [ ] diagnostics: WorkflowDiagnostic[] }).
- [ ] - [ ] Tests cover: precedence (Module → Global → Project → Launch)
- [ ] intra-scope collision
- [ ] missing vars
- [ ] undeclared-but-referenced var (warns)
- [ ] declared-but-unused var (info)
- [ ] extends: inheritance merge.
- [ ] - [ ] END-TO-END INTEGRATION TEST: realistic 3-step workflow fixture under __fixtures__/integration/ with golden-snapshot assertion on composed output. Drift in the snapshot fails CI — this is the canonical guardrail catching unintended changes anywhere in FM-parser → loader → composer → renderer chain.
- [ ] - [ ] RENAME launchHeadless → launchHeadlessSubtask. Add required relaySession argument. RelaySession is a DISCRIMINATED UNION not a maybe-undefined object: type RelaySession = { kind: 'tmux'
- [ ] target: string
- [ ] statusLine: (s: string) => void
- [ ] paneStream: (chunk: string) => void } | { kind: 'test'
- [ ] capture: (line: string) => void }. Test paths construct { kind: 'test'
- [ ] capture: vi.fn() }; production paths construct { kind: 'tmux'
- [ ] ... }. Typechecker rejects callers that skip the relay — enforces the visibility tenet at compile time.
- [ ] - [ ] Surfaces a WorkflowDiagnostic { code: 'info' } size-budget notice when composed total exceeds maxWorkflowChars; goes to visible relay in production
- [ ] to test capture in tests.
- [ ] - [ ] Update settingsPure.ts: maxWorkflowChars default raised from 2000 to 50000.

earchibald:
Final grandchild of [OP-184](https://github.com/earchibald/obsidian-projects/issues/222) — the integration glue that turns OP-194 (renderer + `pluginVarRegistry`), OP-195 (module schema + loader), and OP-196 (workflow file schema + `extends:` + `modelRegistry`) into a single load-and-compose pipeline ready for OP-185 to wire into kickoff.

Closes #235.

## Summary

- **`composeWorkflowPure.ts`** — pure composer. Resolves user-var precedence (Module → Global → Project → Launch), runs OP-194's `renderTemplate` for plugin vars, returns `ComposedPrompt { text, orderedChunks, perVarSourceMap, sizeChars, diagnostics }`. No I/O.
- **`composeWorkflow.ts`** — IO seam: `loadModuleSources(app, opts)` wraps OP-195 `loadModules` + OP-196 `loadWorkflowFile`, reads each module's body via `app.vault.read`, returns `ModuleSourceBundle`. `loadAndComposeWorkflow` is a one-stop convenience.
- **`relaySession.ts`** — visibility-tenet discriminated union `{ kind: "tmux", target, statusLine, paneStream } | { kind: "test", capture }`. Factory helpers `makeTmuxRelay()` / `makeTestRelay()`. Type-level enforcement of OP-181 §"Visibility tenet — every step is observable".
- **`launchHeadless` → `launchHeadlessSubtask` rename** with required `relaySession` arg. Streams stdout/stderr through the relay; emits status-line events at every terminal state. Typechecker rejects callers that omit relaySession.
- **`evaluator.ts` + `main.ts`** updated: evaluator's only production caller now constructs `makeTmuxRelay({ target: "obsidian-plugin:<id>", statusLine: Notice, paneStream: console.log })` — visible surfaces inside Obsidian today; OP-185 will wire a richer streaming pane.
- **`settingsPure.ts`** — `injection.maxWorkflowChars` default raised 2000 → 50000 (OP-181 plan §Risks 1: ~5% of smallest 200k context window). `mergeSettings` preserves user-saved overrides; only fresh installs see the new default.
- **`workflowDiagnostic.ts`** — added `size-budget` to `WorkflowDiagnosticCode` for the info-severity overrun notice.
- **End-to-end integration test** under `__fixtures__/integration/` with a realistic 3-step workflow (kickoff/plan/review) extending a global default workflow. Composed output is asserted against `composed.snapshot.md` — drift fails CI. Re-bake with `npm test -- -u`.

## Diagnostics emitted by the composer

- `unknown-module` (error) — step references a module id with no loaded match.
- `missing-var` (warning) — user var declared somewhere but resolved to no value; or unknown plugin var (re-emitted from `renderTemplate`).
- `malformed-frontmatter` (warning) — undeclared-but-referenced user var.
- `malformed-frontmatter` (info) — declared-but-unused user var.
- `size-budget` (info) — composed prompt exceeds `maxWorkflowChars`.
- `schema-mismatch` (error) — step name not found in workflow.

## Tests

1099 / 1099 passing (was 1058). 41 net-new tests covering precedence, intra-scope-collision visibility, missing/undeclared/unused vars, plugin-var fallthrough, size-budget overrun, legacy kickoff splice, `extends:` inheritance, RelaySession routing, `launchHeadlessSubtask` relay events, loader IO behaviour, and the integration golden snapshot. Pre-existing `iTermPrefs.test.ts` infra-failure reproduces identically on unmodified main.

## Smoke (OP-Test)

- Version bump 0.64.0 → 0.65.0 (minor — additive infrastructure; no user-visible behaviour change yet).
- `dev-sync` clean; plugin reloaded; 39 commands registered; `dev:console` clear of errors.

## Test plan

- [x] `npm test` in `plugins/op-obsidian/` — 1099/1099 (excluding pre-existing iTermPrefs infra failure)
- [x] `node scripts/bump-version.mjs minor` (build asserts main.js > manifest.json)
- [x] `node scripts/dev-sync.mjs` into OP-Test
- [x] `obsidian vault=OP-Test eval` smoke probe — 39 commands registered, no console errors
- [x] Integration test golden snapshot generated and committed
- [ ] Adversarial Copilot review

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.64.0",
+  "version": "0.65.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.64.0",
+  "version": "0.65.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/__fixtures__/integration/_op-workflow.md
+++ b/plugins/op-obsidian/src/__fixtures__/integration/_op-workflow.md
@@ -1,0 +1,12 @@
+---
+type: workflow
+schema: 1
+project: fixture-project
+default_agent: claude
+default_model: opus
+steps:
+  - step: kickoff
+    modules: [branching, version-cadence]
+---
+
+Global default workflow inherited by per-project files via `extends:`.
--- a/plugins/op-obsidian/src/__fixtures__/integration/composed.snapshot.md
+++ b/plugins/op-obsidian/src/__fixtures__/integration/composed.snapshot.md
@@ -1,0 +1,21 @@
+## STEP: kickoff
+
+
+You are working on **OP-FIX-1** (Integration fixture issue) in project `fixture-project` on branch `worktree-OP-FIX-1`. Always create an isolated worktree before making changes — no exceptions, including one-line tweaks. The main checkout may be held by the agent that delegated to you.
+
+
+
+After every change to `plugins/op-obsidian/`, run `node scripts/bump-version.mjs <patch|minor|major>` so the manifest, package, and plugin JSON files move in lockstep. Pick the bump level by judgment: patch = fix, minor = additive, major = breaking. Today is 2026-04-26.
+
+
+## STEP: plan
+
+
+Before implementing, read related code (no more than 10 files), then propose a plan covering: approach, files to touch, tests, risks, and what's deliberately out of scope. Send the plan to @earchibald for sign-off before any commit.
+
+
+## STEP: review
+
+
+Once tests pass and CI is green, request an adversarial Copilot review with concrete pressure-test prompts. After review is addressed, merge with `gh pr merge --squash --delete-branch`. PR: https://github.com/example/repo/pull/42. Issue: OP-FIX-1 (parent: OP-FIX-PARENT).
+
--- a/plugins/op-obsidian/src/__fixtures__/integration/global-modules/branching.md
+++ b/plugins/op-obsidian/src/__fixtures__/integration/global-modules/branching.md
@@ -1,0 +1,9 @@
+---
+id: branching
+title: Always work in a git worktree
+type: workflow-module
+scope: kickoff
+order: 10
+---
+
+You are working on **{{id}}** ({{title}}) in project `{{project}}` on branch `{{branch}}`. Always create an isolated worktree before making changes — no exceptions, including one-line tweaks. The main checkout may be held by the agent that delegated to you.
--- a/plugins/op-obsidian/src/__fixtures__/integration/global-modules/version-cadence.md
+++ b/plugins/op-obsidian/src/__fixtures__/integration/global-modules/version-cadence.md
@@ -1,0 +1,11 @@
+---
+id: version-cadence
+title: Version bump cadence
+type: workflow-module
+scope: kickoff
+order: 20
+vars:
+  - "package_name=op-obsidian"
+---
+
+After every change to `plugins/{{vars.package_name}}/`, run `node scripts/bump-version.mjs <patch|minor|major>` so the manifest, package, and plugin JSON files move in lockstep. Pick the bump level by judgment: patch = fix, minor = additive, major = breaking. Today is {{today}}.
--- a/plugins/op-obsidian/src/__fixtures__/integration/per-project/MODULES/plan-rules.md
+++ b/plugins/op-obsidian/src/__fixtures__/integration/per-project/MODULES/plan-rules.md
@@ -1,0 +1,12 @@
+---
+id: plan-rules
+title: Plan-mode rules for fixture project
+type: workflow-module
+scope: plan
+order: 10
+vars:
+  - "max_files=10"
+  - reviewer_handle
+---
+
+Before implementing, read related code (no more than {{vars.max_files}} files), then propose a plan covering: approach, files to touch, tests, risks, and what's deliberately out of scope. Send the plan to {{vars.reviewer_handle}} for sign-off before any commit.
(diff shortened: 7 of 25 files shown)


earchibald:
@copilot — adversarial review please. Pressure-test prompts:

1. **`renderTemplate` order vs token shape disjointness.** The composer does user-var substitution first, then runs `renderTemplate` for plugin vars. If a user var's resolved value contains a `{{plugin_var}}`-shaped substring (e.g., `vars.title_template = "Issue {{id}}"`), the second pass will resolve the inner token. Is that intended? What if a user wants to keep the literal `{{id}}` text?
2. **`USER_VAR_TOKEN_RE` lookahead with `g` flag.** `extractUserVarRefs` calls `USER_VAR_TOKEN_RE.exec` in a loop and resets `lastIndex` at the start. Is there any reentrancy concern (composer called recursively, two simultaneous walks against the same regex)? Should we `new RegExp` per call instead?
3. **`resolveUserVar` walks Module → Global → Project → Launch with later wins.** The function uses bare `if` chain, not `switch`. If we add a fifth precedence layer later, will the chain stay correct? Why not a `for (const layer of LAYERS)` loop with explicit precedence ordering?
4. **`resolveUserVar` with two modules declaring the same `name` at the same scope.** OP-195's intra-scope-collision validator already flagged it, but the composer breaks the tie alphabetically by module id. Is that defensible? What if a user reorders the workflow's `modules:` list expecting first-listed to win?
5. **`renderModule`'s declared-but-unused diagnostic.** Walks `module.vars` against `extractUserVarRefs(body)` of the *original* body. If a module declares `vars: [foo, bar]` and the body has `{{vars.foo}} {{vars.bar}}`, but `bar` resolves to a string that itself looks like `{{vars.baz}}` (which is also declared), does `bar` count as "used"? Should the second-pass body affect the unused calculation?
6. **`composeWorkflow` finalisation `text.length` vs JS UTF-16 surrogate pairs.** The size-budget cap measures `text.length` (UTF-16 code units). A module body full of emoji could over-count or under-count vs displayed characters. Is that the right unit? Should it be byte length, grapheme count, or token estimate?
7. **`launchHeadlessSubtask` relay status-line coverage.** The new code emits `[status]` events at: launch, timeout, exit≠0, parse error, spawn error, completion. Are all "terminal" states covered? What about an explicit kill from a parent process (SIGINT mid-stream)? Does the relay still close the loop?
8. **`launchHeadlessSubtask` partial-chunk emission to relay.** Each `data` event on stdout/stderr emits one `[stream]` chunk. If `claude -p` streams a partial JSON byte-by-byte, the relay sees chunked partial garbage. Is that acceptable for a "visible" relay, or should we line-buffer?
9. **`makeTmuxRelay` factory misnaming.** The variant is called `tmux` but the only production callsite uses `Notice` + `console.log` (no tmux pane). Should the variant be renamed to a more general `surface`, or is the historical naming worth keeping for OP-185 forward-compat?
10. **`composeWorkflow.ts:loadModuleSources` body-read race.** Between `loadModules`'s metadataCache walk and the `getAbstractFileByPath` body read, a file could move/rename. The code emits `schema-mismatch` and skips. Is that the right diagnostic code (vs. `unknown-module`)? Does the composer downstream handle a partially-loaded module set correctly?
11. **Snapshot test fragility.** `composed.snapshot.md` was bake-generated by the test itself on first run. If a contributor unintentionally changes a module body, they'll see the snapshot diff. But what if a transitive dependency change (e.g., `renderTemplate` regex tweak) shifts the snapshot — does the test failure point them to the right file to inspect?
12. **`integration.test.ts` hand-mirrored frontmatter.** The fixture's frontmatter is duplicated between the `.md` files (for human readability) and the `FIXTURE_FILES` array (for the fakeApp). If someone updates the `.md` frontmatter without updating the array, the test will silently use the array. Should we YAML-parse the fixture instead?
13. **`evaluator.ts` relaySession plumbing in `EvaluatorFlowDeps`.** The new field is required, breaking any non-test caller. Are there callers outside `main.ts:runEvaluatorForIssue` that need to be updated?
14. **`workflowDiagnostic.ts:size-budget` code addition.** The exhaustiveness helper `assertNeverCode` will break any consumer that pattern-matches over the union. Do any existing surfaces switch on the code without a `default`?
15. **`maxWorkflowChars` migration.** Bump from 2000 → 50000. `mergeSettings` preserves user-saved values, but the OP-149 settings UI may have a clamp / range validator that now mis-renders. Any UI surfaces that need updating?

Thanks!

PR #260, opened 2026-04-26
OP-201: Settings — top-level Workflows group + unified WorkflowDiagnostic formatter

### LINKED ISSUE #239: 3a: Settings — new top-level Workflows group + empty-state + diagnostic formatter
Tracked as op issue **OP-201**.

## Scope

- [ ] - [ ] Parent: OP-186 (Child 3 umbrella)
- [ ] - [ ] Add a NEW TOP-LEVEL 'Workflows' group ABOVE the Advanced collapsibles (not as a 9th Advanced collapsible — too important and too referenced to bury). Inside the group: Workflow modules list (from loadModuleSources)
- [ ] per-module status badges
- [ ] missing-var diagnostics
- [ ] 'Explain workflow' action
- [ ] 'Available variables' reference panel sourced from pluginVarRegistry.
- [ ] - [ ] Empty-state: when no modules are present
- [ ] show 'No modules yet — open the 5-min Quickstart' linking to docs (Child 10a / OP-?-10a) plus a one-click 'Install example library from _op-modules/_examples/' button so a brand-new user has a path from zero.
- [ ] - [ ] Implements the unified WorkflowDiagnostic formatter — every diagnostic surface (here
- [ ] op-explain-workflow
- [ ] op-list-vars
- [ ] editor squiggles
- [ ] dry-run banner
- [ ] launch-modal pre-flight) renders WorkflowDiagnostic[] through this one formatter. Use full canonical scope names ('Module default'
- [ ] 'Global default'
- [ ] 'Project default'
- [ ] 'Launch override') in user-visible text; abbreviations (M/G/P/L) appear only in compact tooltip text
- [ ] never in primary copy.
- [ ] - [ ] DOM probe smoke tests asserting the new group renders
- [ ] empty-state shows when no modules exist
- [ ] diagnostic formatter renders all WorkflowDiagnostic codes correctly.
- [ ] - [ ] The legacy 'Workflow inline cap (chars)' row in the Injection collapsible is retired at cutover (Child 8a / OP-191-8a) — not in this issue.

earchibald:
## Summary

- Adds a NEW top-level `Workflows` group in op-obsidian Settings (between Daily and Advanced — too important and too referenced to bury under Advanced) showing the workflow modules tree, per-module severity badges, formatter-rendered inline diagnostics, an `Explain workflow` action, and an `Available variables` panel sourced from `pluginVarRegistry`.
- Empty-state with a one-click `Install example library` button that writes 2 starter modules into `Projects/_op-modules/` (idempotent; never overwrites). Closes the zero-to-working gap a brand-new user faces.
- Ships the unified **`WorkflowDiagnostic` formatter** (`workflowDiagnosticFormat.ts`) that every other 3* surface (op-explain-workflow, op-list-vars, editor squiggles, dry-run banner, launch-modal pre-flight) will consume. Renders canonical precedence-scope names (`Module default` / `Global default` / `Project default` / `Launch override`) in primary copy; abbreviations (M/G/P/L) only in compact tooltip text per the OP-201 spec. Exhaustive `switch` over `WorkflowDiagnosticCode` so a new code at the type level forces a deliberate decision here.

Implements OP-201 (3a of OP-186 umbrella). Closes earchibald/obsidian-projects#239.

## Test plan

- [x] `npx vitest run` — 1105 / 1105 pass (47 new)
- [x] `npx tsc --noEmit` — 8 errors, all pre-existing in `main` (verified by `git stash && tsc`); zero new errors introduced
- [x] OP-Test smoke probes (CLAUDE.md §5 recipe, `vault=OP-Test` on every call):
  - `.op-settings__group--workflows` + `[data-op-section="workflows"]` render
  - `[data-op-empty-state="true"]` shows when no modules exist
  - After clicking `Install example library`: 2 modules render with `ok` badges + Explain buttons; empty-state gone
  - Synthetic malformed module surfaces in the orphan-diagnostic footer with formatter prose, severity badge `E`, and canonical `Malformed frontmatter` tooltip (NOT abbreviation in primary copy)
  - `Available variables` panel renders all 16 `PLUGIN_VAR_REGISTRY` entries
  - Existing Daily group (5 rows) and Advanced group (8 collapsibles) intact, no regressions
  - `dev:console` clean, no errors

## Out of scope

Own issues under OP-186: `op-explain-workflow` / `op-list-vars` CLIs (OP-203), launch-modal pre-flight (OP-204), bad-model recovery dialog (OP-205), dry-run banner (OP-206), editor squiggles (OP-207), retiring legacy injection cap (OP-208).

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.65.0",
+  "version": "0.66.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.65.0",
+  "version": "0.66.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/settings.ts
+++ b/plugins/op-obsidian/src/settings.ts
@@ -576,6 +576,13 @@ export class OpSettingsTab extends PluginSettingTab {
               if (r.installed.length) parts.push(`${r.installed.length} installed`);
               if (r.skipped.length) parts.push(`${r.skipped.length} skipped (already present)`);
               notify(`Example library: ${parts.join(", ") || "no changes"}.`, 6000);
+              // metadataCache populates each new file's frontmatter
+              // asynchronously after vault.create resolves — rendering
+              // immediately can miss files whose cache hasn't caught up,
+              // showing only the first installed module. Wait for the
+              // batch-resolve event (or a short timeout fallback) before
+              // re-rendering so loadModules sees every installed file.
+              await waitForCacheSettle(this.app, 500);
               this.rerenderSection("workflows");
             } catch (e) {
               notify(`Could not install example library: ${(e as Error).message}`, 8000);
@@ -1935,6 +1942,32 @@ function appendBadge(
   parentEl.createSpan({
     cls: `op-workflow-badge op-workflow-badge--${severity}`,
     text: `${count} ${severity}${count === 1 ? "" : "s"}`,
+  });
+}
+
+/**
+ * Wait for `metadataCache` to fire its batch-`resolved` event (signalling
+ * the cache has caught up with newly-created files), or for a timeout
+ * fallback — whichever comes first. Without this, a render that follows a
+ * `vault.create` can miss files whose frontmatter hasn't been parsed yet
+ * (the file is in `getMarkdownFiles()` but `getFileCache(file).frontmatter`
+ * is still null). The timeout is the floor: if the cache was already
+ * resolved before we registered the listener, we still rerender promptly.
+ */
+async function waitForCacheSettle(app: App, timeoutMs: number): Promise<void> {
+  await new Promise<void>((resolve) => {
+    let done = false;
+    const settle = (): void => {
+      if (done) return;
+      done = true;
+      app.metadataCache.offref(ref);
+      window.clearTimeout(handle);
+      resolve();
+    };
+    // The 'resolved' event fires after the cache has finished indexing the
+    // current batch of changes — exactly the signal we need.
+    const ref = app.metadataCache.on("resolved", settle);
+    const handle = window.setTimeout(settle, timeoutMs);
   });
 }
 
--- a/plugins/op-obsidian/src/settings.ts
+++ b/plugins/op-obsidian/src/settings.ts
@@ -6,6 +6,15 @@ import {
   prepareFuzzySearch,
 } from "obsidian";
 import { notify } from "./notificationLog";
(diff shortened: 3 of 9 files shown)


earchibald:
@copilot — adversarial review please. Concrete pressure-test prompts:

**Unified formatter contract (`workflowDiagnosticFormat.ts`)**
1. The formatter is the public contract every other 3* surface (OP-203/204/206/207) will consume. Are the `FormattedDiagnostic` fields stable enough that adding a new diagnostic code in the future won't force a breaking change to the shape? Suggest fields you'd add now to forestall future v2 churn.
2. The OP-201 spec says abbreviations (`M`/`G`/`P`/`L`) MUST NOT appear in primary copy. Walk every code path that produces user-visible text (`diagnosticToLine`, `diagnosticToBlock`, `formatDiagnostic` consumers in `settings.ts`) and find any place an abbreviation could leak — including badge `aria-label`/`title`, modal headers, `extra.precedenceScope` rendering, and the orphan-footer message column.
3. `extra.precedenceScope` is read with `isPrecedenceScope` and a runtime check. What happens if a 1c/1d diagnostic emits a typo (e.g. `"Module"`/`"PROJECT"` instead of `"module"`/`"project"`)? Is the silent-drop behavior the right call, or should it surface an internal-error diagnostic?
4. The exhaustiveness fence in `formatDiagnostic` lists every code as a fallthrough that returns the same shape. Is there a more idiomatic exhaustiveness pattern that would also force authors to handle new codes' code-specific extras (e.g. `import-collision.scope` once 1c lands)?

**Empty-state + example library (`workflowExamples.ts`)**
5. `installExampleLibrary` writes via `app.vault.adapter.write`, bypassing `app.vault.create`. Does this register the new files with `metadataCache` correctly so a subsequent `loadModules` call sees them? Check the OP-Test smoke trace if needed (commit shows `modulesNow:2` after install, so it works in practice — but is there a race or first-call edge case the smoke missed?).
6. The `_op-modules/` mkdir uses adapter primitives directly — are there permission/casing edge cases on Windows or iCloud-backed vaults that'd silently fail?
7. Empty-state Quickstart link points at `docs/quickstart.md` which doesn't exist yet (ships in OP-?-10a). External link with `target=_blank` and `rel=noopener` — do we want to render it disabled until the doc lands, or is the live-but-404 approach OK?

**Settings UI (`settings.ts`)**
8. The new top-level group is added to `display()` but not wired into the existing fuzzy search. The `Available variables` panel rows are `.op-workflow-vars-panel__row`, not `.setting-item`, so they aren't filterable. Is that the right call, or should the search find variable tokens too?
9. The `WorkflowExplainModal` opens diagnostic blocks as `<pre>`. Multi-line content has variable-length first lines — does this layout break on long messages or non-ASCII chars (e.g. CJK in `vars[].description`)?
10. `loadModules(this.app)` runs synchronously inside `display()` on every settings-tab open. For a vault with hundreds of modules and large frontmatter, is this fast enough, or do we need debouncing / a cache?
11. The orphan-diagnostic footer renders any diagnostic whose `moduleId` doesn't appear in `modules`. Could a future code emit diagnostics that legitimately lack a `moduleId` and SHOULDN'T be in this footer (e.g. a workflow-file-level error)? Consider whether an explicit `sourceKind` field would be cleaner.

**General**
12. Concurrency: two settings tabs open at once (multiple Obsidian windows on the same vault) → both call `loadModules`, both render. Any state that could collide?
13. Schema drift: the `WorkflowDiagnostic.extra` shape is documented but not type-enforced per code. If 1c starts emitting `import-collision` with `extra.scope` as a string, will the formatter's current defensive coding (`extra?.path` typed-check) extend cleanly?

PR #261, opened 2026-04-26
OP-198: workflowMode flag + kickoff composer splice (2a)

earchibald:
## Summary

- Adds top-level `workflowMode` (`'legacy' | 'modules'`, defaults to `'legacy'`) and `workflowVars` (`Record<string, string>`, defaults to `{}`) fields to `OpSettings`. `mergeSettings` validates both and falls back to defaults on unknown literals or non-string entries — pre-OP-198 `data.json` blobs load cleanly with default values.
- `buildPrompt` branches on `workflowMode === 'modules'` to splice OP-197's `loadAndComposeWorkflow` output into the kickoff prompt in place of the legacy `WORKFLOW.md` inline-blob. Legacy code path is preserved byte-identical (covered by a literal-snapshot regression test). Per-mode and per-step launch-context wiring lands in OP-199 (2b) and OP-200 (2c).
- `openAgent` threads `settings.workflowMode` + `settings.workflowVars` into the existing `buildPrompt` call. No behavior change for default users.

Issue: OP-198 — child of OP-185 (Child 2 umbrella).

## Test plan

- [x] `vitest run` — 1111 / 1111 passing (5 new settings cases + 6 new `buildPrompt` cases). The single failing suite `iTermPrefs.test.ts` is the same pre-existing infra issue documented across OP-194/195/196/197.
- [x] Build green at op-obsidian 0.66.0 (minor — additive settings + opt-in code path).
- [x] dev-sync into OP-Test → live probe confirms `manifest.version === '0.66.0'`, `settings.workflowMode === 'legacy'`, `settings.workflowVars === {}` (mergeSettings supplied defaults since persisted `data.json` carries no key).
- [x] Settings tab still renders (5 daily rows + 8 advanced sections per the OP-164 layout); console clean after reload.
- [x] In-memory mutation to `'modules'` + `{ reviewer_handle: '@me' }` succeeds without runtime error; restored by `plugin:reload` from disk.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.65.0",
+  "version": "0.66.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.65.0",
+  "version": "0.66.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/openAgent.ts
+++ b/plugins/op-obsidian/src/openAgent.ts
@@ -127,6 +127,8 @@ export async function openAgent(
     injection: settings.injection,
     vaultBasePath,
     mode,
+    workflowMode: settings.workflowMode,
+    workflowVars: settings.workflowVars,
   });
 
   const { scriptPath, tmuxSession, tmuxWindow } = await launchInTerminal({
--- a/plugins/op-obsidian/src/promptBuild.test.ts
+++ b/plugins/op-obsidian/src/promptBuild.test.ts
@@ -456,6 +456,95 @@ describe("buildPrompt — OP-198 (2a)", () => {
     expect(out).toContain("Issue: OP-1 — Demo issue");
   });
 
+  it("modules mode: empty kickoff step (modules:[]) suppresses section — does NOT fall back to legacy", async () => {
+    // WORKFLOW.md exists, step is found, but `modules: []` — the composer
+    // returns an empty ComposedPrompt (text: ''). The caller must suppress
+    // the section without injecting the legacy inline blob. A user who opts
+    // into modules mode with an empty kickoff step is explicitly requesting
+    // no workflow injection at that step — silently getting the old inline
+    // blob would be surprising.
+    const app = fakeApp([
+      {
+        path: "Projects/demo/WORKFLOW.md",
+        frontmatter: {
+          type: "workflow",
+          schema: 1,
+          project: "demo",
+          default_agent: "claude",
+          default_model: "opus",
+          steps: [{ step: "kickoff", modules: [] }],
+        },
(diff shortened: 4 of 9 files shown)


earchibald:
@copilot please pressure-test this PR. Specific concerns I'd like you to probe:

1. **Migration safety for `mergeSettings`** — does an existing `data.json` blob from a pre-OP-198 install (no `workflowMode`, no `workflowVars` keys) load with `'legacy'` and `{}` defaults? Does an explicitly stored `workflowMode: 'modules'` survive a round-trip? What about a persisted `workflowVars` map with a mix of valid string entries and (defensively) non-string entries that a third-party tool could have written?

2. **Byte-identical legacy regression** — the `REFERENCE_LEGACY` constant in `promptBuild.test.ts` locks in the legacy output. Is the literal-snapshot approach robust to whitespace edge cases (trailing newline on `WORKFLOW.md` body, missing trailing newline, BOM, CRLF)? Would the test catch a one-character drift in the `## Project workflow` header or the `From Projects/...:` introducer line?

3. **`composeKickoffSection` fallback semantics** — when the composer returns `composed === null` *or* `composed.text.trim() === ''`, the caller falls through to the legacy block. Is this actually the right behaviour for an empty-step modular workflow (`steps: [{ step: "kickoff", modules: [] }]`)? Could a user who explicitly opted into modules be surprised by silently getting the legacy inline blob back? What about a non-empty step whose only module renders to whitespace after var substitution?

4. **`RenderContext` partial fill** — `buildKickoffRenderContext` sets `parent: null`, `branch/repo_path/model: undefined`. Modules that reference `{{branch}}`, `{{model}}`, or `{{repo_path}}` will emit `missing-var` warnings via `composed.diagnostics`. The diagnostics never surface in the prompt itself in 2a — is there any caller path (logging, telemetry) where they'd leak unexpectedly? Is the `today` derivation via `new Date().toISOString().slice(0, 10)` timezone-safe (UTC vs local), and does it matter if a module renders `{{today}}` and the user's timezone disagrees with UTC at midnight?

5. **`workflowVars` validation in `mergeSettings`** — the per-entry filter drops non-string keys (impossible from a JSON object, but defensively) and non-string values. Does it correctly handle `undefined` values, `null` values, the empty string `""`, very long strings, and string keys that contain dots/slashes (which would tokenise oddly inside `{{vars.<name>}}`)?

6. **`BuildPromptArgs` backward-compat** — `workflowMode?` and `workflowVars?` are both optional. Are there any other in-repo callers besides `openAgent.ts:124`? (I see only the one — `grep -rln "buildPrompt" plugins/op-obsidian/src/` returns just `promptBuild.ts` and `openAgent.ts`.)

7. **Settings tab rendering** — no UI surface for the new fields was added (deferred to OP-186). Confirm the settings tab still renders cleanly with the new top-level fields present in the settings object — the OP-164 layout iterates over a known set of section ids, but a new top-level field shouldn't break it.

8. **`includeWorkflow: false` interaction** — when both `workflowMode: 'modules'` AND `includeWorkflow: false`, the workflow branch is entirely skipped (gate works on both code paths). Is this correct, or should `includeWorkflow` be deprecated / repurposed as a "legacy on/off" toggle now that `workflowMode` exists?

9. **`stripFrontmatter` duplication** — `composeWorkflow.ts:stripFrontmatter`, `workflowFilePure.ts:stripWorkflowFrontmatter`, and `promptBuild.ts:stripFrontmatter` are byte-for-byte equivalent (per OP-197 §"Strip a markdown file's frontmatter fence"). Is this duplication a maintenance risk, or is the runtime-dependency justification solid?

10. **Composer's `step: 'kickoff'` hardcode** — is this the right step name to use at every kickoff site? Will OP-199 (2b) need to override it per launch mode (e.g., `step: 'plan'`, `step: 'review'`)? Should this be parameterised now to avoid a follow-up signature change?

Find anything I missed.

PR #262, opened 2026-04-26
OP-199: per-mode workflow injection on every launch (2b)

earchibald:
## Summary

- Threads the active `AgentLaunchMode` through to OP-197's modular workflow composer at every `openAgent` and evaluator launch, so each mode composes its own step instead of always `kickoff`.
- Broadens `RenderContext` with launch-context (`branch`, `repo_path`, `parent`) so modules referencing those vars resolve cleanly at launch.
- Lenient kickoff fallback keeps single-`kickoff` workflow files working unchanged; OP-198's explicit-suppress contract (step exists with `modules: []`) is preserved.

## What's new

- `BuildPromptArgs` grows optional `repoPath`, `branch`, `parentId` (backward-compat at every existing callsite).
- `composeWorkflowSection` exported from `promptBuild.ts` so non-`buildPrompt` callers (the evaluator today; future per-step subtasks) can inject the same workflow section.
- `modeToWorkflowStep(mode)` helper next to `normalizeMode` in `agentProfiles.ts`.
- `gitBranch.ts` — fail-soft `git rev-parse --abbrev-ref HEAD` IO helper with an injectable `execFn` test seam.
- `EvaluatorFlowDeps.composeWorkflowSection?` — `main.ts` wires it for `workflowMode === 'modules'` with `workflowStep: 'evaluate'`. Fail-soft if the composer throws.

## Tests

- 18 new (gitBranch=6, evaluator=4, promptBuild=8). Per-mode parametrised `it.each` over `evaluate`/`plan`/`implement`/`review`/`finalize`. Lenient-fallback + explicit-suppress tests lock in the OP-198 / OP-199 boundary.
- Full suite: **1189 / 1189**. Pre-existing `iTermPrefs.test.ts` infra-failure unchanged.

## Version

`op-obsidian 0.66.0 → 0.67.0` (minor — additive: new optional args + exported helper + evaluator dep).

## Test plan

- [x] Unit tests cover each mode → step mapping, lenient fallback, explicit-suppress, launch-context render, parent sentinel.
- [x] Build + dev-sync to OP-Test; plugin reloaded clean (39 commands, settings tab renders, console clean).
- [ ] Modules-mode end-to-end exercise pending OP-189 (workflow file migration) and OP-186 (settings UI surface) — out of scope here per the OP-198 successor plan.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.66.0",
+  "version": "0.67.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.66.0",
+  "version": "0.67.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/agentProfiles.ts
+++ b/plugins/op-obsidian/src/agentProfiles.ts
@@ -151,6 +151,16 @@ export function normalizeMode(mode: AgentLaunchMode): Exclude<AgentLaunchMode, "
  * `normalizeMode + stringify` recipe — and so a future reshuffle (e.g.
  * folding `'finalize'` into `'review'`) is one edit, not a vault-wide
  * find-and-replace.
+ *
+ * **Pre-OP-185 `'work'` steps.** `normalizeMode('work') === 'implement'`, so
+ * a workflow file that has `step: 'work'` (the pre-OP-185 name) will never
+ * match — OP-199 always requests `'implement'`. When the workflow *also*
+ * lacks an `'implement'` step, the lenient kickoff fallback in
+ * `composeWorkflowSection` fires and uses `'kickoff'` instead. When the
+ * workflow has neither `'implement'` nor `'kickoff'`, the section is
+ * suppressed entirely. The fix is to rename the step in the workflow file
+ * (`step: 'work'` → `step: 'implement'`). Deliberately not honoured here to
+ * avoid ambiguity with an `'implement'` step that may coexist in the same file.
  */
 export function modeToWorkflowStep(mode: AgentLaunchMode): string {
   return normalizeMode(mode);
--- a/plugins/op-obsidian/src/agentProfiles.ts
+++ b/plugins/op-obsidian/src/agentProfiles.ts
@@ -142,6 +142,20 @@ export function normalizeMode(mode: AgentLaunchMode): Exclude<AgentLaunchMode, "
   return mode === "work" ? "implement" : mode;
 }
 
+/**
+ * OP-199 (2b): map an `AgentLaunchMode` to the workflow-file step name to
+ * compose for that launch. Today the mapping is the identity over the
+ * normalized mode (`work → implement`), so each non-deprecated mode picks the
+ * step whose `step:` literal matches its name. Encapsulated here so callers
+ * who want to derive the step from the mode never duplicate the
+ * `normalizeMode + stringify` recipe — and so a future reshuffle (e.g.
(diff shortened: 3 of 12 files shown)


earchibald:
@copilot — adversarial pressure-test for OP-199 (per-mode workflow injection + launch-context plumbing). Please push back hard on these specific prompts; I'd rather find a real bug now than ship a regression.

**1. Lenient kickoff fallback — diagnostic detection.** `composeWorkflowSection` in `promptBuild.ts` detects "step missing entirely" via `composed.diagnostics.some(d => d.code === "schema-mismatch" && d.stepId === requestedStep)`. Are there code paths that emit a `schema-mismatch` diagnostic with `stepId === requestedStep` for reasons OTHER than "step is absent from `workflow.steps[]`" (e.g. a malformed module body, a per-step model validation error)? If so, the fallback could silently take over when the user actually wanted explicit-suppress for some other reason. Walk every emit-site in `composeWorkflowPure.ts` / `workflowFilePure.ts` / `loadWorkflowFile` and confirm only the step-not-found path matches.

**2. Lenient fallback — empty-step preservation.** The OP-198 contract is "step exists with `modules: []` → no section, no kickoff bleed-in". My test `explicit-suppress preserved` covers `modules: []` but does it cover other zero-output paths — e.g. a step whose modules ALL render to whitespace after var substitution, or a step pointing at a module id that fails to load (`unknown-module` diagnostic)? Does the fallback fire in those cases? Should it?

**3. `gitBranchAt` race conditions.** `git rev-parse --abbrev-ref HEAD` is invoked at every `openAgent` launch. What happens if the user is mid-rebase/mid-merge (HEAD is a temporary ref)? What about a fresh `git init` repo with no commits (HEAD is unborn)? What about a worktree whose underlying HEAD file is being torn down between two parallel launches? Document the failure mode for each case.

**4. `parentId` from metadataCache.** `readParentId` in `openAgent.ts` calls `app.metadataCache.getFileCache(file)`. If the cache hasn't yet observed the issue note (e.g. a freshly-created issue, OR an ENOENT race when the file is being moved by `op-resolve` mid-launch), it returns `null`. Is that the right default, or should the launch block until the cache populates? What if `parent:` is a list (`parent: [OP-1, OP-2]`) instead of a scalar — my code returns `null` defensively, but is silent dropping correct, or should we render the first / surface a diagnostic?

**5. `composeWorkflowSection` export — public surface.** It's now exported from `promptBuild.ts`. Are there hidden invariants in `BuildPromptArgs` that callers (the evaluator today; future callsites) might violate? E.g. is it safe to pass an `entry` whose `path` doesn't exist on disk? An `injection` with `maxWorkflowChars: 0`? A `vaultBasePath` of `""`?

**6. Evaluator prompt size.** Modules-mode prepends a composed `## Project workflow\n\n…` to the evaluator's prompt. The composer's `maxWorkflowChars` is 50000 by default — that's a lot to hand to a `claude -p` invocation that's supposed to do a quick triage. Is there a path where a long workflow file pushes the evaluator into a different reasoning regime (truncation, slower turnaround, COMPLEXITY trailer parse failure)? Should the evaluator clamp harder?

**7. `evaluator.ts` — `composeWorkflowSection` deps shape.** `EvaluatorFlowDeps.composeWorkflowSection?: () => Promise<string | null>`. The dep is fully closure-captured at the `main.ts` callsite (binds `app`, `settings`, `entry`, etc.). If `entry` mutates between dep construction and dep invocation (e.g. status flips from `open` to `in-progress` mid-flow), does the closure see the post-mutation state, and does that matter for the composed render context?

**8. `modeToWorkflowStep` — `'work'` legacy alias.** `normalizeMode('work') === 'implement'`, so `modeToWorkflowStep('work')` returns `'implement'`. If a workflow author has a step named `'work'` from a pre-OP-185 setup, it would never match. Is that the right call (force-migrate) or should we honor `'work'` if explicitly defined? Today's behavior: lenient kickoff fallback would catch them — but only if their workflow ALSO lacks `'implement'`.

**9. Backward-compat at every `buildPrompt` callsite.** I confirmed only `openAgent.ts:124` calls `buildPrompt` in production. Any URI handler / external CLI handler / test helper / future `launchHeadlessSubtask` that spawns from the same plugin? `git grep "buildPrompt(" plugins/op-obsidian/` should reveal them.

**10. Test coverage symmetry.** Each mode in the `it.each` covers happy-path composition. None covers what happens when `repoPath: undefined` (meta-only project) for that mode — the launch context degrades but the step composition itself should still work. Worth adding, or covered by other tests?

Please address by severity, push fixes as commits if you find anything actionable, and document design rationale for prompts where you think the current implementation is right.

PR #263, opened 2026-04-26
OP-203: op-explain-workflow + op-list-vars diagnostic CLIs

earchibald:
## Summary

- New `op-explain-workflow issue=<ID> mode=<m> [agent=<a>]` — composes the prompt the launcher would produce + per-var precedence breakdown (Module / Global / Project / Launch default — full canonical names, never abbreviations) + every diagnostic rendered through the OP-201 formatter.
- New `op-list-vars [project=<slug>] [issue=<ID>]` — every `PLUGIN_VAR_REGISTRY` entry as JSON, augmented with current resolved value when an issue is provided.
- Pure layer (`explainWorkflowPure.ts`, `listVarsPure.ts`) owns JSON shape; IO seam (`explainWorkflow.ts`) owns vault reads. Both verbs write structured payloads to `Projects/_scratch/op-last-response.md` per the existing convention.

## Why

Closes OP-203 (3c grandchild of OP-186). 3a (OP-201) shipped the Settings panel + unified diagnostic formatter; 3c surfaces the same underlying composer state through the CLI, so an agent debugging a launch sees the same prose the editor squiggles and the dry-run banner show. By construction `composed.text` is byte-identical to what the launcher would inject — both surfaces share `composeWorkflow`.

## Implementation notes

- Per-var rows source from `ComposedPrompt.perVarSourceMap` rather than the diagnostic stream — diagnostics carry `precedenceScope` only on resolution failures, but every referenced var deserves a row regardless of outcome.
- Declared-but-unreferenced vars surface through the existing `malformed-frontmatter` info diagnostic, not a synthesized row, to keep the formatter contract the only source of "things to surface".
- Full canonical names everywhere in primary output (per OP-201 contract); abbreviations only present in the structured payload as the separate `scopeAbbrev` field for tooltip-only consumers.

Bumps `op-obsidian` 0.66.0 → 0.67.0 (minor — additive new public surface).

## Test plan

- [x] 21 new pure tests (golden snapshots covering happy path, all four precedence layers, shadow-precedence, missing-var, registry-only / issue-bound list-vars).
- [x] Full vitest suite green (1161 pass / 0 fail).
- [x] `tsc --noEmit` — 8 pre-existing errors unchanged from `main` baseline; zero new errors introduced.
- [x] Smoke-tested on OP-Test (vault=OP-Test) — `op-list-vars` registry-only, `op-list-vars issue=DM-1`, and `op-explain-workflow issue=DM-1 mode=kickoff` against a synthetic 1-module then 2-module workflow. Verified `op-last-response.md` shape, canonical scope labels, and the missing-var diagnostic surfacing through the unified formatter.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.67.0",
+  "version": "0.68.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package-lock.json
+++ b/plugins/op-obsidian/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "op-obsidian",
-  "version": "0.60.1",
+  "version": "0.67.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "op-obsidian",
-      "version": "0.60.1",
+      "version": "0.67.0",
       "license": "MIT",
       "dependencies": {
         "protobufjs": "^8.0.1"
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.67.0",
+  "version": "0.68.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/cliHandlers.test.ts
+++ b/plugins/op-obsidian/src/cliHandlers.test.ts
@@ -12,6 +12,8 @@ import {
   parseGetWorkflowParams,
   parseEditWorkflowParams,
   parseGetSkillParams,
+  parseExplainWorkflowParams,
+  parseListVarsParams,
 } from "./cliHandlers";
 
 describe("parseWorkParams", () => {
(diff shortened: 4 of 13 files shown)


earchibald:
@copilot please review with these pressure-test prompts:

1. **Param-validation drift between CLI and URI handlers.** `cliHandlers.ts:parseExplainWorkflowParams` rejects whitespace-only/embedded-whitespace `agent=`; `uriHandlers.ts:handleOpExplainWorkflowUri` re-implements a similar check. Are they in lockstep, or does one accept input the other rejects (e.g. tab characters, NBSP, leading-only whitespace)? Same question for `parseListVarsParams` vs `handleOpListVarsUri`.

2. **Mode → step coupling.** `explainWorkflow.ts` passes `args.mode` as `composeWorkflow.step` verbatim. What happens when `mode` is a value the workflow file doesn't declare? Verify the diagnostic surfaces correctly through the formatter (vs. erroring out the CLI).

3. **`launchVars` is hardcoded to `{}` in `explainWorkflow.ts`.** That's correct for now (OP-204 ships the launch override panel) but worth checking: does the CLI's silence about launchVars match the dry-run banner's behavior when no override is active? If they diverge, the "verbatim with dry-run" claim breaks.

4. **`vars` array sorting.** `buildExplainPayload` sorts by name for stable golden snapshots. `composeWorkflow.perVarSourceMap` walks a `Set<string>` — does this introduce any non-determinism elsewhere (e.g. diagnostic order)? Check that the sort doesn't mask a real ordering bug.

5. **Project-vars source.** `readProjectVars(app, project)` reads `STATUS.md` frontmatter `vars:` map. What if `vars:` exists but is `null`, an array, a list of `{name: x, value: y}` objects, or contains non-string values? Does it silently drop or surface a diagnostic?

6. **AgentId coercion.** `resolveProfileById` falls back to `settings.defaultAgent` when the supplied `agent` doesn't match a known `AgentId`. Should this instead emit an error diagnostic so the user sees `agent=copilotcc` (typo) vs. an unconditional fallback to claude?

7. **`op-list-vars project=<slug>` without an issue.** The current behavior synthesizes `{ project: <slug> }` as the only context — every other field is null. Confirm this is intentional vs. a degraded mode that should warn.

8. **Schema/contract symmetry with the dry-run preview (OP-206).** The acceptance criterion is "behavior identical to dry-run preview's composed string." OP-206 isn't shipped yet — does the CLI's payload shape allow the future dry-run to consume it without renaming fields?

PR #264, opened 2026-04-26
OP-202: op-edit-module + {{ autocomplete + keyed-map affordance

earchibald:
## Summary

Ships OP-186 child 3b — the workflow-authoring surface trio:

- **`op-edit-module`** palette command + URI + CLI launcher. Mirrors `op-edit-workflow` but pointed at a single module file (`Projects/_op-modules/<id>.md` or `Projects/<slug>/MODULES/<id>.md`). Picker modal lists every loaded module plus "+ New global / project module…" entries; `NewModuleIdModal` collects the id for new files. Agent prompt (`editModulePure.ts`) inlines the module schema reference so authors don't have to recall YAML syntax.
- **Rebranded `op-edit-workflow` prompt** — `buildEditWorkflowPrompt` now embeds the modern schema (`type: workflow`, `schema: 1`, list-or-keyed-map `default_agent` / `default_model`, optional `extends:`, `steps[]`). Fresh-file authoring lands in the right shape without round-tripping to the docs.
- **`{{` autocomplete + vars-block snippet** via Obsidian's `EditorSuggest` API. Pure trigger detection + candidate union in `varSuggest.ts`; Obsidian-aware seam in `varSuggestObsidian.ts`. Activates only on workflow modules and per-project `WORKFLOW.md`. Suggestions union `PLUGIN_VAR_REGISTRY` (15 always-on vars) with user vars from the current module + global + project modules, source-labelled with OP-201's canonical `Plugin (Launch override)` / `Module default` / `Global default` / `Project default` strings. Empty bullet under a `vars:` block opens the object-form snippet `{ name: NAME, default: "" }` with the cursor selecting `NAME`.
- **Keyed-map model affordance** — `op: switch workflow model: to per-agent map` palette command, gated to multi-agent workflows whose `default_model` is still scalar/list. Rewrite goes through `fileManager.processFrontMatter` so only `default_model` is touched (rest of the file is bit-stable).
- Version bump 0.66.0 → 0.67.0 (minor — additive surfaces; no breaking changes).

## Test plan

- [x] vitest: 1214 pass on baseline + 1252 pass with this PR (new: 32 in `varSuggest.test.ts`, 6 in `editModulePure.test.ts`, 5 in `cliHandlers.test.ts`, 1 in `editWorkflow.test.ts`).
- [x] tsc: 8 pre-existing baseline errors unchanged; 0 new errors introduced.
- [x] OP-Test smoke (per CLAUDE.md §smoke):
  - `obsidian help` lists `op-edit-module` alongside `op-edit-workflow`.
  - Palette commands `op: edit workflow module` and `op: switch workflow model: to per-agent map` are registered.
  - On a synthesised global module file, `EditorSuggest.onTrigger` returns null on issue notes, fires on the module, and returns 21 candidates at `{{` (15 plugin + 6 user across pre-existing OP-Test fixtures + the new module's `tone`).
  - Empty bullet under `vars:` returns the object-form snippet with cursor offset 8.
  - Keyed-map command on `default_agent: [claude, gemini] / default_model: opus` rewrites to `default_model: { claude: opus, gemini: opus }`.
  - `dev:console` clean (no errors, no warnings).

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.68.0",
+  "version": "0.67.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.67.0",
+  "version": "0.69.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.67.0",
+  "version": "0.69.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.68.0",
+  "version": "0.67.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/cliHandlers.test.ts
+++ b/plugins/op-obsidian/src/cliHandlers.test.ts
@@ -11,6 +11,7 @@ import {
   parseScaffoldParams,
   parseGetWorkflowParams,
   parseEditWorkflowParams,
+  parseEditModuleParams,
   parseGetSkillParams,
   parseExplainWorkflowParams,
   parseListVarsParams,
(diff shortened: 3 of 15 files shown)


earchibald:
@copilot please review this PR. Specific pressure tests:

1. **`getDoubleBraceTrigger` boundary cases.** Cursor immediately after `{{`, mid-token, after `{{vars.`, between `{{` and `}}` of an existing token, on a line with both `{{` and `}}`. Does the 64-char scan window fail-safe on long single-line YAML? What happens with `{{ foo }}` (whitespace inside)? With `{{vars.foo.bar}}` (chained dot)? The current code rejects these — verify the rejection path doesn't leak a stale `pending` from a prior trigger.

2. **`getVarsBlockTrigger` walker.** The walker stops at the first strictly-shallower line and checks for `vars:`. What if `vars:` is INSIDE another mapping (e.g. nested config under `agents:`)? What if the YAML has `vars: [foo]` inline (flow style) — does our `^vars\s*:\s*$` regex correctly NOT trigger then? What about a 4-space-indented bullet under a 2-space-indented `vars:` (deep nesting)? Confirm the walker handles all these without crashing or emitting wrong-snippet triggers.

3. **`buildCandidates` precedence + de-dup.** Three modules declare a var named `tone` — global `orient`, project `house-style`, current module being edited. The de-dup keeps project-scope (highest in the layered walk). Verify: (a) the module currently being edited DOES appear in the candidate list when its var name doesn't clash with anything higher, and (b) when the current module shadows a global, the global's entry is dropped (we only ship the topmost). Pay attention to `loadVarSuggestContext` in `varSuggestObsidian.ts` where it filters the current module out of the layered lists — is the path-equality filter robust to `normalizePath` differences?

4. **Snippet insertion + `setSelection` cursor.** The vars-block snippet inserts `{ name: NAME, default: "" }` and then `setSelection`s `NAME`. Does this work correctly when the bullet line had trailing whitespace (we trim down to `- ` before inserting)? When the bullet column is non-zero (deep nesting)? Does the editor's undo merge correctly so a single Cmd+Z reverts both the insert and the selection?

5. **`activeKeyedMapCandidate` gate.** The current implementation reads `frontmatter.default_agent` + `default_model` from `metadataCache`. metadataCache is async-populated; the gate can return null right after a file edit even when the FS state would qualify. Is the user surprise (palette command "missing" briefly) acceptable, or should we fall back to reading the file directly? Also: does `readModelScalarOrList` return `[]` (qualifying!) for `default_model: ""` (empty string)? That would make us "convert" an already-empty value into per-agent empties — is that desirable?

6. **`fileManager.processFrontMatter` rewrite atomicity.** If the frontmatter contains list-form `default_agent` (`- claude\n- gemini`) and we rewrite `default_model` to an object, does Obsidian preserve the list form or normalize? Verify the diff on the smoke-test target — the worktree shows the list got reformatted to the YAML default; is the rest of the frontmatter (project, schema, type, extends) bit-stable?

7. **CLI param parsing.** `parseEditModuleParams` defaults `scope=project` when `--project` is supplied, and `scope=global` otherwise. What if the user passes `module=foo project=demo scope=global`? Currently we accept and use `scope=global` (project arg is silently ignored as a working-dir hint). Is that the right call, or should we warn?

8. **Schema-reference fidelity in the rebranded prompt.** The inline schema lists per-agent keyed-map as opt-in via the palette command. Does the prompt accurately describe when an agent should NOT introduce the keyed-map form on a fresh file? (Default authoring should stay scalar/list per the issue's acceptance criteria.)

PR #265, opened 2026-04-26
OP-200: per-step agent/model resolver + typo-vs-overflow rule + recovery hook (2c)

earchibald:
## Summary

The 2c slice of OP-185 (Child 2 umbrella). Final piece — 2a (OP-198) shipped the `workflowMode` flag + kickoff splice, 2b (OP-199) shipped per-mode injection on every launch + launch-context plumbing. 2c lands the **per-step agent/model resolver**, the **typo-vs-cross-agent overflow rule**, and the **recovery-hook seam** for Child 3e.

## What ships

- **`stepResolver.ts`** (new pure module) — `resolveStepAgentAndModel({ step, workflow, detection, fallbackAgent })`. Walks the step's `agent:` list with safe-failure semantics (uninstalled agents are silently skipped); resolves `model:` from either flat-list (`kind: "all"`) or per-agent keyed-map (`kind: "perAgent"`). Returns `{ agent, canonicalModel, agentSource, modelSource }` or throws.
- **`BadModelSpecError` / `NoInstalledAgentError`** — structured failures carrying per-entry classification so the recovery dialog can render "you wrote 'opuss' (typo) and 'pro' (gemini's, not claude's)".
- **Typo-vs-overflow rule (precise)** — each entry classified once: `valid`, `cross-agent-overflow` (known to ANY other agent's registry), `typo` (unknown everywhere). **Typo precedence**: any typo → throw immediately, even when a later entry would have validated. Never silently substitutes.
- **`recoveryDialog.ts`** (new stub seam) — `openRecoveryDialog({ app, issueId, error })` opens the op-obsidian Settings tab and surfaces a sticky Notice describing the failing entry. OP-?-3e replaces the body with the interactive picker without touching any callsite.
- **`openAgent.ts` wiring** — loads `WORKFLOW.md`, finds the step record for `modeToWorkflowStep(mode)`, runs the resolver. On success: resolver's chosen agent overrides `settings.defaultAgent`, canonical model id is appended to claude's launch flags as `--model X`, `RenderContext.model` populates. On failure: actionable Notice with 'Open recovery dialog now' action; launch is cancelled.
- **`promptBuild.ts`** — new `BuildPromptArgs.resolvedModel?: string`; `buildLaunchRenderContext` reads it into `RenderContext.model`. The `{{model}}` slot 2b left undefined finally resolves cleanly when the workflow declares per-step overrides.

## Visibility-tenet compliance

The visibility tenet is enforced at the type level by `launchHeadlessSubtask`'s required `relaySession` field (already there from OP-197). 2c does **not** introduce a new headless subtask path — the resolver itself is pure. What 2c adds is the **failure surface**: instead of an unhandled throw burying the recovery affordance in console.error, the actionable Notice + recovery seam ensures auto-advance failures don't dead-end the chain.

## Tests

- `stepResolver.test.ts` (new, **27 cases**):
  - Agent walking: first installed wins; first listed wins among multiple installed; `workflow.defaultAgent` fallback; empty-list `fallbackAgent`; `NoInstalledAgentError` when none installed; defensive unknown-agent-id handling.
  - Typo-vs-overflow: alias resolution, versioned-id passthrough, walk past overflow, **typo precedence even when a later entry would validate**, **typo precedence with overflow + valid + typo mix**, `BadModelSpec` payload contents.
  - Keyed-map: chosen agent's key lookup, multi-value walk, missing-key returns `undefined` (not an error), per-agent typo throws, sibling-entry typos for OTHER agents are ignored.
  - Defaults inheritance: workflow defaults inherited when step is silent, step overrides workflow, no spec anywhere → `undefined`.
- `recoveryDialog.test.ts` (new, **5 cases**) — settings tab opens, sticky-Notice phrasing per failure mode (overflow / typo / no-installed-agent), defensive when `app.setting` is unavailable.

Full suite: **1261 / 1261** (was 1195 pre-OP-200; +66 from this work). Pre-existing `iTermPrefs.test.ts` infra-failure unchanged from the OP-194/195/196/197/198/199 baseline.

## Smoke

OP-Test vault, post `dev-sync`:
- `manifest.version` = `0.70.0` ✓
- 41 commands registered ✓
- Settings tab renders (5 daily rows + 8 advanced sections — matches OP-164 layout) ✓
- Console clean after reload ✓
- Plugin instance reachable; `settings.defaultAgent` and `settings.workflowMode` defaults intact ✓

## Pressure-test prompts for adversarial review (Copilot)

1. **Typo precedence walk** — confirm `model: ["opuss", "opus"]` throws even though `opus` would have validated. Walk all `validateModelName` emit sites to verify the typo classification can't be silently masked.
2. **Cross-registry collision** — what if a model name appears in BOTH the chosen agent's registry AND another agent's? Today: validates as `valid` for chosen agent, never reaches the overflow check. Confirm this is correct (registry overlap is benign — chosen agent wins).
3. **Per-agent map with `kind: "all"` defaults** — if step's `model:` is a per-agent map but the chosen agent has no key, is the workflow's `defaultModel` consulted as a second-level fallback? Today: NO. The step's `model:` shadows the workflow default entirely; missing key → `undefined`. Confirm this is the intended scoping.
4. **Resolver bypass via picker** — when the user opens the picker (`forcePick` or `alwaysPick`), the resolver is skipped (`args.agentOverride` short-circuits). Is this the right precedence? Workflow file is a default, manual pick wins.
5. **`WORKFLOW.md` partial state** — file exists, schema valid, but workflow's `defaultAgent: []` and step has no `agent:` either. Resolver returns `fallback` source. Confirm no surprising behavior.
6. **Auto-advance + resolver error** — `advanceFlowAndLaunch` calls `setFlow(... nextFlow)` BEFORE `doOpenAgent`. When the launch fails on a `BadModelSpecError`, the flow has advanced but no agent ran. Is this acceptable, or should the resolver run earlier (peek the workflow file pre-`setFlow`)?
7. **`--model` flag injection is claude-only** — gemini and copilot are second-class; their CLIs never see the resolved model on the command line. Modules referencing `{{model}}` still resolve in their prompt. Confirm this divergence is documented.
8. **Recovery dialog seam stability** — the `openRecoveryDialog({ app, issueId, error })` signature is what 3e replaces. Does the seam carry enough context (issue id + structured error)? The error class hierarchy (`BadModelSpecError | NoInstalledAgentError`) — are there other resolver failure modes 3e will need to handle?
9. **`BadModelSpec.bad` for `<defaults>` step** — when the resolver fails on workflow defaults (no step record), `stepId` is `<defaults>`. Confirm downstream (Notice text, recovery dialog) renders this correctly without crashes.
10. **Idempotent re-runs** — calling the resolver twice with the same input should yield the same output. Walk the `MODEL_REGISTRY` lookups for any `Math.random()` / iteration-order surprises (`Object.keys` order is well-defined for non-integer keys, but worth a sanity check).

## Test plan

- [ ] Full vitest suite passes (1261 / 1261 confirmed locally)
- [ ] Smoke at OP-Test (settings tab renders, console clean, version flips)
- [ ] Adversarial Copilot review pressure-tests above

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.69.0",
+  "version": "0.70.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.69.0",
+  "version": "0.70.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/openAgent.ts
+++ b/plugins/op-obsidian/src/openAgent.ts
@@ -89,16 +89,23 @@ export async function openAgent(
   const detection = detector.get() ?? (await detector.refresh());
   const mode: AgentLaunchMode = args.mode ?? "work";
 
-  // OP-200 (2c): per-step resolver. Runs only when the user did NOT manually
-  // override the agent (manual picker wins over the workflow file's
-  // declarations — workflow declarations are a default, not a hard
-  // constraint). The resolver also fires when args.agentOverride is unset
-  // and the workflow file declares an agent the user's settings.defaultAgent
-  // doesn't match. On a BadModelSpecError / NoInstalledAgentError the launch
-  // is cancelled and the user sees an actionable Notice with a recovery seam.
+  // OP-200 (2c): per-step resolver. Skipped when the user has explicitly
+  // indicated they want to choose the agent themselves — workflow declarations
+  // are a default, not a hard constraint. There are three bypass conditions:
+  //   • args.agentOverride — caller already nominated a specific agent (e.g.
+  //     a URI-dispatch with agent= or an auto-advance with a fixed agent id).
+  //   • args.forcePick — the invoking command requested "always show the
+  //     picker" for this launch. Honouring the user's intent means the
+  //     workflow file must not pre-empt the modal.
+  //   • settings.alwaysPick — user-level preference; same rationale as
+  //     forcePick.
+  // When any bypass fires, resolved = null and pickAgent handles agent
+  // selection normally (modal or default — pickAgent checks mustPick itself).
+  // Model resolution is also skipped on bypass because the model spec is
+  // bound to the workflow's declared agent; a user-chosen agent may differ.
   let resolved: ResolveStepOutput | null;
   try {
-    resolved = args.agentOverride
+    resolved = args.agentOverride || args.forcePick || settings.alwaysPick
       ? null
       : await loadAndResolveStep(app, args.entry.project, mode, detection, settings.defaultAgent);
   } catch (err) {
--- a/plugins/op-obsidian/src/openAgent.ts
+++ b/plugins/op-obsidian/src/openAgent.ts
@@ -17,10 +17,19 @@ import { gitBranchAt } from "./gitBranch";
 import { workIssue } from "./workIssue";
 import { resolveWorkingDir } from "./workingDir";
 import { launchInTerminal } from "./terminalLaunch";
-import type { AgentDetector } from "./agentDetect";
+import type { AgentDetector, DetectionMap } from "./agentDetect";
 import { AgentPickerModal } from "./modals";
 import { userError } from "./userError";
 import { iTermDefaultsDomainPresent, showITermTmuxPrefsNotice } from "./iTermPrefs";
+import { loadWorkflowFile } from "./workflowFile";
+import {
+  BadModelSpecError,
+  NoInstalledAgentError,
+  resolveStepAgentAndModel,
+  type ResolveStepOutput,
+} from "./stepResolver";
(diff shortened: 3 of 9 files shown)


earchibald:
@copilot please pressure-test this PR with the 10 prompts in the description. Specific focus areas:

1. **Typo precedence walk** — confirm `model: ["opuss", "opus"]` throws `BadModelSpecError(reason: "typo")` even though `opus` would have validated. Verify that no `validateModelName` emit site can silently mask a typo.
2. **Cross-registry collision** — what if a model name appears in BOTH the chosen agent's registry AND another agent's? Today: validates as `valid` for chosen, never reaches overflow check. Right behavior?
3. **Per-agent map with fallback** — step's `model:` is a per-agent map but chosen agent has no key. Today: returns `undefined` (no error). Should the workflow's `defaultModel` be consulted as a second-level fallback? Document the scoping.
4. **Resolver bypass via picker** — manual picker (`forcePick` / `alwaysPick`) skips the resolver. Right precedence?
5. **`WORKFLOW.md` partial state** — `defaultAgent: []` and step has no `agent:` either. Resolver returns `fallback`. Surprises?
6. **Auto-advance + resolver error** — `setFlow` happens BEFORE `doOpenAgent`. Resolver throws → flow advanced but no agent. Acceptable, or should resolver run pre-`setFlow`?
7. **`--model` flag injection is claude-only** — gemini/copilot CLIs never see it. Modules referencing `{{model}}` still resolve in prompt. Document divergence?
8. **Recovery dialog seam stability** — `openRecoveryDialog({ app, issueId, error })` is what 3e replaces. Does the seam carry enough context? Other resolver failure modes 3e needs?
9. **`BadModelSpec.bad` for `<defaults>` step** — `stepId === "<defaults>"`. Notice text and recovery dialog handle this correctly?
10. **Idempotent re-runs** — same resolver input, same output. Walk the `MODEL_REGISTRY` lookups for iteration-order surprises.

PR #266, opened 2026-04-26
OP-204: launch-time variable override panel (3d)

earchibald:
## Summary

- New `LaunchAgentModal` with a folded **Workflow variables** disclosure on the launch surfaces — per-row badges use the canonical scope names (`Module default` / `Global default` / `Project default` / `Launch override`) per OP-201's contract, single-letter abbreviations only in tooltips. Reset-to-default per row + show-all toggle (only-referenced by default).
- `launchVars: Record<string, string>` plumbed through `OpenAgentArgs` → `BuildPromptArgs` → `composeWorkflowSection` so the composer's already-supported Launch precedence layer (level 4) finally sees real values.
- `obsidian://op-open-agent` URI now accepts `var.<name>=<value>` keys (or the packed `vars=name=val\\nname2=val2` form) — `parseLaunchVarsFromUri` extracts both shapes.
- Auto-advance carry-through via `fm.launch_vars`: `openAgent` persists non-empty maps to the issue's frontmatter; `advanceFlowAndLaunch` re-threads them on the next-stage launch; `op-resolve` clears the field symmetric with `agent:`.
- 0.69.0 → 0.70.0 (minor — additive: new modal, new URI param, new frontmatter field, no breakage).

## Adversarial review

1. **Empty-string Launch override** — passing `var.tone=` (empty value) MUST set the row's currentScope to `launch` with `currentValue: ""` and shadow lower layers. Confirm `composeWorkflowPure.resolveUserVar` and `varOverridePanelPure.buildPanelRows` agree on this (both use `Object.prototype.hasOwnProperty.call` rather than truthiness).
2. **Module-default tie-break** — when two modules declare the same var with different defaults, the lowest-id module wins. Confirm `firstModuleDefault` walks an already-sorted-by-id array and that the panel's pick matches `composeWorkflowPure.resolveUserVar` byte-for-byte (intra-scope-collision diagnostic is OP-195's; we must not double-handle).
3. **Frontmatter coercion** — `readLaunchVarsFromFrontmatter` stringifies numbers/booleans but drops nested objects/arrays/dates. If a user hand-edits `launch_vars: { count: 3 }` in YAML, `count` should arrive as `"3"` at the composer (which only accepts strings). Confirm the metadataCache hand-back doesn't smuggle in a weirdly-typed field.
4. **Modal carry-through edge case** — opening the modal on an issue that already has `fm.launch_vars` should pre-populate the inputs *and* flip the badge to `Launch override` immediately. Verify the initial render uses `initialLaunchVars` (not the empty default).
5. **Auto-advance schema sprawl** — `advanceFlowAndLaunch` writes `fm.launch_vars` once per stage. Multi-step flows should not duplicate / merge / lose entries between stages. Worth confirming the metadataCache freshness assumptions hold (we yield once with `setTimeout(0)` before reading the refreshed entry).
6. **URI packed-form last-wins** — `?var.tone=A&vars=tone=B` should resolve to `{tone: "B"}` (packed wins on duplicates). Confirm the prefix scan happens before the `collectRepeated` walk so the order is right.
7. **Modal cancel symmetry** — Cancel returns `undefined` and `doOpenAgent` returns early without touching `fm.launch_vars`. Confirm we don't write the field on the modal's "decided=false" path.
8. **`workflowMode === "legacy"`** — the panel renders an inert hint instead of empty rows. Launches still proceed with `launchVars: {}` (no-op at the composer). Confirm we don't crash on a project with no `WORKFLOW.md`.

## Test plan

- [x] 22 vitest cases on `varOverridePanelPure.ts` — precedence layers, tie-break, empty-string override, frontmatter coercion, merge/clear immutability.
- [x] 9 vitest cases on `parseLaunchVarsFromUri` — both forms, malformed names, packed-vs-prefix precedence.
- [x] Full vitest suite: 1261 / 1261 pass.
- [x] `tsc --noEmit`: 8 pre-existing baseline errors unchanged; 0 new errors.
- [x] OP-Test smoke: modal opens via `op-open-agent-pick` against `demo` project's 2-module workflow, panel summary reads "2 referenced", expand reveals correct rows, badge flips Module default → Launch override on input with the right canonical-name + tooltip-abbrev shape, console clean.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "OP-204",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {}
+}
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.70.0",
+  "version": "0.71.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.70.0",
+  "version": "0.71.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/launchAgentModal.ts
+++ b/plugins/op-obsidian/src/launchAgentModal.ts
@@ -1,0 +1,410 @@
+import { App, Modal, Setting } from "obsidian";
+import type { AgentId } from "./agentProfiles";
+import { AGENT_IDS } from "./agentProfiles";
+import { loadAndComposeWorkflow, type LoadedModule } from "./composeWorkflow";
+import type { ComposedPrompt } from "./composeWorkflowPure";
+import type { OpSettings } from "./settings";
+import {
+  buildPanelRows,
+  clearLaunchOverride,
+  mergeLaunchOverride,
+  type PanelRow,
+} from "./varOverridePanelPure";
+import { precedenceScopeAbbrev, precedenceScopeLabel } from "./workflowDiagnosticFormat";
+
+// OP-204 (3d): launch-modal IO seam. Wraps the agent picker AND the folded
+// "Workflow variables" disclosure on a single Obsidian Modal. The pure data
+// (`varOverridePanelPure.buildPanelRows`) drives row rendering; this module
+// owns DOM writes, state management, and the final `{ agentId, launchVars }`
+// payload that `doOpenAgent` plumbs into `openAgent`.
+//
+// Surfaces that currently route through here: the `op-open-agent-pick`
+// palette command, the sidebar shift-click affordance, and any future
+// "with vars…" entry point. The silent default-agent path stays silent —
+// users explicitly opt into the modal for the vars surface.
+
+export interface LaunchAgentModalArgs {
+  /** Project slug — used to load modules and resolve the workflow. */
+  project: string;
+  /** Agents installed on this machine — populates the dropdown. */
+  installed: AgentId[];
+  /** Default agent to preselect. */
+  defaultAgent: AgentId;
+  /** Vault settings — `workflowMode` decides whether modules even apply. */
+  settings: OpSettings;
+  /** Existing launch overrides (e.g. carried from `fm.launch_vars`). */
+  initialLaunchVars?: Record<string, string>;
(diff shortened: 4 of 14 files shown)


earchibald:
@copilot — please pressure-test this PR end-to-end. Concrete prompts:

1. **Empty-string Launch override**: passing `var.tone=` (empty value) MUST set the row's currentScope to `launch` with `currentValue: ""` and shadow lower layers. Verify `composeWorkflowPure.resolveUserVar` and `varOverridePanelPure.buildPanelRows` agree byte-for-byte on this (both should use `Object.prototype.hasOwnProperty.call` rather than truthiness).

2. **Module-default tie-break**: when two modules declare the same var with different defaults, the lowest-id module wins. Confirm `firstModuleDefault` walks an already-sorted-by-id array and that the panel's choice matches the composer's choice. The intra-scope-collision diagnostic is OP-195's responsibility — we must not double-handle.

3. **Frontmatter coercion**: `readLaunchVarsFromFrontmatter` stringifies numbers/booleans but drops nested objects/arrays/dates. If a user hand-edits `launch_vars: { count: 3 }` in YAML, `count` should arrive at the composer as `"3"` (string) — confirm the metadataCache hand-back doesn't smuggle in a weirdly-typed field that breaks the composer's `Record<string,string>` assumption.

4. **Modal carry-through edge case**: opening the modal on an issue that already has `fm.launch_vars` should pre-populate the inputs *and* flip the badge to `Launch override` immediately. Verify the initial render uses `initialLaunchVars` (not the empty default), and that `expanded: args.startExpanded ?? Object.keys(this.launchVars).length > 0` correctly auto-expands when overrides exist.

5. **Auto-advance schema sprawl**: `advanceFlowAndLaunch` writes `fm.launch_vars` once per stage via `openAgent`. Multi-step flows should not duplicate / merge / lose entries between stages. Worth confirming the metadataCache freshness assumptions hold — we yield once with `setTimeout(0)` before reading the refreshed entry.

6. **URI packed-form last-wins**: `?var.tone=A&vars=tone=B` should resolve to `{tone: "B"}` (packed wins on duplicates). Confirm the prefix scan happens **before** the `collectRepeated` walk in `parseLaunchVarsFromUri` so the order is right; the existing test asserts this but worth eyeballing.

7. **Modal cancel symmetry**: Cancel returns `undefined` and `doOpenAgent` returns early without touching `fm.launch_vars`. Confirm we don't write the field on the modal's `decided=false` path.

8. **`workflowMode === "legacy"`**: the panel renders an inert hint instead of empty rows. Launches still proceed with `launchVars: {}` (no-op at the composer). Confirm we don't crash on a project with no `WORKFLOW.md` and that the modal doesn't try to load modules in legacy mode.

9. **Modal Enter-key submit**: `scope.register([], "Enter", ...)` submits when no input is focused — but inputs swallow Enter via the early-return on `t.tagName === "INPUT"`. Verify a user typing into a row input and pressing Enter does NOT submit the modal mid-keystroke.

10. **CSS class symmetry**: `wrap.toggleClass(...)` calls in `updateRowBadge` should toggle both `--override` and `--unset` correctly when the row transitions between states (e.g. Set → Reset to default → typed Empty string).

Issue #243, opened 2026-04-26
3e: Bad-model-spec recovery dialog with diff/.bak/revert


earchibald:
Tracked as op issue **OP-205**.

## Scope

- [ ] - [ ] Parent: OP-186 (Child 3 umbrella)
- [ ] - [ ] When the resolver (2c / OP-200) throws BadModelSpecError on an interactive launch — OR when a headless auto-advance failure surfaces a Notice with 'Open recovery dialog now' and the user clicks it — render this modal:
- [ ] - [ ] Modal lists offending step.id and bad name; shows allowed-models for the resolved agent in two side-by-side columns (aliases + versioned ids)
- [ ] pulled live from modelRegistry.
- [ ] - [ ] Lets user pick a valid model via typeahead/dropdown; launch proceeds with the pick as a Launch-override (workflow file unchanged).
- [ ] - [ ] SECOND ACTION: 'Patch workflow with this model'. Real safety net required:
- [ ] - [ ]   • Show a UNIFIED DIFF of the about-to-be-patched line(s) before the user confirms.
- [ ] - [ ]   • Write a timestamped backup at <workflow-file>.bak-<YYYYMMDD-HHmmss> BEFORE mutating; dialog displays the backup path post-patch.
- [ ] - [ ]   • 'Revert last patch' affordance (palette command + dialog button) restores from the most recent .bak-* and trashes it; one-step undo only
- [ ] older patches require manual restore.
- [ ] - [ ] Cancellation aborts the launch (no silent fallback).
- [ ] - [ ] Same dialog reachable from op-explain-workflow's warnings list when validation surfaces bad-model warnings ahead of an actual launch — so users can fix the workflow file proactively
- [ ] not just at-launch.
- [ ] - [ ] DOM probe smoke tests for both the dialog and the revert flow.

Issue #10, opened 2026-04-22
sidebar should display github issue if configured


earchibald:
Tracked as op issue **OP-40**.

PR #267, opened 2026-04-26
OP-205: bad-model recovery dialog with diff/.bak/revert (3e)

earchibald:
## Summary

Replaces the OP-200 recovery-dialog stub with the real interactive modal called for in OP-186 child 3e (issue [#243](https://github.com/earchibald/obsidian-projects/issues/243)).

- **Interactive launch failure** → `BadModelSpecError` opens the modal directly. The user picks a valid model from the registry (typeahead + side-by-side `aliases` / `versioned` columns rendered live from `MODEL_REGISTRY`). Two actions:
  - `Use as launch override` — the launch retries with `launchModelOverride: <canonical>` set; workflow file is left untouched. The override is validated against the chosen agent's registry and re-throws on mismatch (re-opens the dialog) so we never silently substitute.
  - `Patch workflow with this model` — runs the pure planner, opens a unified-diff confirm sub-modal, and on confirm writes a `.bak-<YYYYMMDD-HHmmss>` backup BEFORE mutating the workflow file. Single-occurrence enforcement: ambiguous matches disable the patch button rather than risk rewriting the wrong copy.
- **Headless auto-advance failure** keeps the OP-200 actionable Notice with `Open recovery dialog now`. New `OpenAgentArgs.interactive` (default `true`) flips to `false` from `advanceFlowAndLaunch` so the user isn't yanked into a modal they didn't trigger.
- **Revert** — palette command `op: revert last workflow patch` + button in the dialog footer. Restores from the most recent `.bak-*` and trashes it; older backups untouched (one-step undo only).
- **Advisory entrypoint** — palette command `op: open recovery dialog for project workflow` opens the same modal proactively for any project whose `validateWorkflowModels` returned `bad-model` warnings, so users can fix the file before the next launch.
- **Cancellation** aborts the in-flight launch (no silent fallback).

## Files

- new: `recoveryPatch.{ts,test.ts}` (pure planner + diff + timestamp + latest-backup picker — 21 vitest cases), `recoveryPatchApply.{ts,test.ts}` (vault-side controllers — 6 cases against a fake `VaultLike`)
- rewritten: `recoveryDialog.{ts,test.ts}` — real `RecoveryDialogModal` + diff confirm sub-modal + revert row; `describeFailure` and `synthesizeBadModelErrorFromDiagnostic` exported for the headless Notice + advisory entrypoints
- edited: `openAgent.ts` (`launchModelOverride` + `interactive` + `onResolverFailureResolved` on `OpenAgentArgs`; resolver short-circuit on override; interactive vs headless surface), `main.ts` (`doOpenAgent` retry plumbing + `advanceFlowAndLaunch interactive=false` + two new palette commands), `styles.css` (`.op-recovery-modal__*` classes)
- bump: `0.71.0 → 0.72.0` (minor — additive: new commands + new optional `OpenAgentArgs` fields)

## Tests

- vitest **1385 / 1385** (35 new: 21 pure-planner + 6 vault-controller + 8 modal phrasing/diagnostic synth)
- live OP-Test DOM smoke (executed pre-PR):
  1. `runOpenRecoveryDialogForProject("demo")` after seeding `default_model: opuss` → modal renders with 3 alias chips + 7 versioned chips, `h2 == "Recover: bad model spec — Projects/demo"`.
  2. Click `opus` chip → picker resolves to `claude-opus-4-7`, patch button enabled, status reads "Resolved → claude-opus-4-7. Patch will rewrite opuss → claude-opus-4-7 in Projects/demo/WORKFLOW.md."
  3. Click Patch → confirm sub-modal renders the unified diff verbatim → Apply → `default_model: claude-opus-4-7` on disk + `Projects/demo/WORKFLOW.md.bak-<ts>` exists.
  4. `runRevertWorkflowPatchForProject("demo")` → file restored to `default_model: opuss`, `.bak-*` trashed.

## Test plan

- [ ] Reviewer pulls the branch, runs `npm test` in `plugins/op-obsidian/` → 1385/1385.
- [ ] Reviewer seeds a bad-model in any project's `WORKFLOW.md` (e.g. `default_model: opuss`), runs `op: open recovery dialog for project workflow`, verifies the modal columns + picker resolve.
- [ ] Reviewer exercises both actions: `Use as launch override` (no file change), `Patch workflow with this model` (file change + `.bak-*` sibling).
- [ ] Reviewer runs `op: revert last workflow patch` → verifies the file is restored from the backup and the backup is gone.
- [ ] Adversarial Copilot review (see follow-up PR comment).

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.71.0",
+  "version": "0.72.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.71.0",
+  "version": "0.72.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/main.ts
+++ b/plugins/op-obsidian/src/main.ts
@@ -116,6 +116,12 @@ import { openLaunchAgentModal } from "./launchAgentModal";
 import { editWorkflow } from "./editWorkflow";
 import { editModule } from "./editModule";
 import { loadModules } from "./workflowModule";
+import { loadWorkflowFile } from "./workflowFile";
+import {
+  openRecoveryDialog,
+  synthesizeBadModelErrorFromDiagnostic,
+} from "./recoveryDialog";
+import { revertLastWorkflowPatch, type VaultLike, type VaultFileLike } from "./recoveryPatchApply";
 import { VarEditorSuggest } from "./varSuggestObsidian";
 import { readAgentList, readModelScalarOrList } from "./varSuggest";
 import { launchInTerminal, SHARED_TMUX_SESSION, tmuxWindowName } from "./terminalLaunch";
@@ -722,6 +728,25 @@ export default class OpPlugin extends Plugin {
       id: "op-edit-module",
       name: "op: edit workflow module",
       callback: () => this.runEditModuleCommand(),
+    });
+
+    // OP-205 (3e): proactive recovery surface — open the dialog without an
+    // in-flight launch. Resolves a project, reads its WORKFLOW.md, finds the
+    // first bad-model diagnostic, and renders the same modal in advisory
+    // mode so users can fix the file before the next launch.
+    this.addCommand({
+      id: "op-open-recovery-dialog",
+      name: "op: open recovery dialog for project workflow",
+      callback: () => void this.runOpenRecoveryDialogCommand(),
+    });
+
+    // OP-205 (3e): one-step undo of the most recent .bak-* for a workflow
+    // file. Reusable from anywhere; a button in the dialog dispatches the
+    // same command.
+    this.addCommand({
+      id: "op-revert-workflow-patch",
+      name: "op: revert last workflow patch",
+      callback: () => void this.runRevertWorkflowPatchCommand(),
     });
 
     this.addCommand({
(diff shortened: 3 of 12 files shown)


earchibald:
@copilot please pressure-test this PR. Focus on these concrete failure modes:

1. **Single-occurrence enforcement edge cases.** `planBadModelPatch` rejects multi-match cases as `ambiguous`. What if the bad name appears in **prose inside a body comment of YAML** (e.g. `default_model: opuss  # was opuss in a prior commit`)? My rule scans only frontmatter via `extractFrontmatter`, so the inline-comment "opuss" inside the same line counts. Is that the right call, or should I narrow the match to value-only positions?

2. **Frontmatter terminator ambiguity.** `extractFrontmatter` splits on `\r?\n---(?:\r?\n|$)` to find the closing fence. What if the workflow file has a YAML scalar containing `---` on its own line (e.g. inside a multi-line `description: |`)? Could that prematurely terminate the frontmatter and skew the patch range?

3. **Backup-write/modify race.** `applyBadModelPatch` writes the `.bak-<ts>` via `vault.create` BEFORE `vault.modify`. If `vault.modify` throws, we leak a `.bak-<ts>` of the unmodified file (harmless but visible in the user's vault). Worse case: if the user runs the dialog twice within the same second, both attempts hit the same `formatBackupTimestamp` and the second `vault.create` throws `EEXIST`. Should I add a 3-digit counter suffix, or fall back to retry-with-bumped-timestamp?

4. **Revert when latest .bak's contents diverge from the workflow's pre-patch state.** A user could hand-edit `WORKFLOW.md` after a patch landed. The `Revert last patch` then writes the OLD pre-patch contents over the new hand-edits — silently. Is the warranty here ("one-step undo of the most recent patch") clear enough, or do we need a confirm step that diffs current vs. backup before clobbering?

5. **`launchModelOverride` validation throw → re-open loop.** When the resolver throws on an invalid override, my code synthesises a fresh `BadModelSpecError` and re-opens the dialog. If the user keeps picking invalid entries (somehow — the picker only offers registry-valid names), is there a recursion risk in the `onResolverFailureResolved` retry path?

6. **Interactive flag default in non-`doOpenAgent` callers.** My `surfaceResolverError` defaults to interactive when the field is undefined. Is there ANY launch path that goes through `openAgent` without going through `doOpenAgent`? (Direct `openAgent` calls from `editWorkflow` / `editModule`?) If so, those callers never set `interactive: false` and would yank the user into a modal during a hook-triggered launch. Audit needed.

7. **Diagnostic-to-error synth heuristic.** `synthesizeBadModelErrorFromDiagnostic` classifies as `typo` when the bad name is unknown to every agent's registry, else as `all-overflow`. What if the registry has overlapping aliases between agents (a hypothetical "opus" appearing in both claude AND another agent's aliases)? `detectOverflowAgent` returns the FIRST hit by `Object.entries` order — non-deterministic if the order ever changes.

8. **Concurrent dialog instances.** `RecoveryDialogModal.onClose` fires `onResolved({kind: "cancelled"})` if `this.resolved` is false, even when the modal closed because a sibling dialog opened. Could a rapid open-then-cancel cause a stale "cancelled" outcome to fire AFTER a successful `override` from a parallel instance?

9. **`vault.trash` argument shape.** I pass `system: true` to `vault.trash` to land the backup in the OS trash, not the vault's `.trash/` folder. If the user has `core-plugin: trash` configured to "permanently delete," the backup is unrecoverable. Is `system: true` the right default, or should this be settings-controlled?

10. **`.bak-*` accumulation.** I never garbage-collect older backups. After many recovery cycles, `Projects/<slug>/` could accumulate a wall of `WORKFLOW.md.bak-*` siblings. Should there be a settings-controlled retention (e.g. keep last 5)?

The dialog's contract is documented in `plugins/op-obsidian/src/recoveryDialog.ts` and the patch planner in `plugins/op-obsidian/src/recoveryPatch.ts`. End-to-end smoke (modal open → pick → patch → revert) was DOM-probed against `OP-Test` pre-PR.

earchibald:
Reviewed Copilot's fix commit `2c323b7` and pulled it into the branch. Evaluation:

**Addressed by Copilot's commit (verified locally + via OP-Test DOM smoke):**
- **#1 (inline-comment ambiguity)**: Documented the conservative-by-design behavior with a new test. The dialog's "ambiguous" branch already disabled the patch button — Copilot's test pins this contract.
- **#2 (frontmatter terminator)**: New test confirms the `\r?\n---(?:\r?\n|$)` regex's `^---` constraint correctly skips an indented `---` inside a YAML literal-block scalar. No code change needed; the test is the contract.
- **#3 (EEXIST race)**: Real fix in `applyBadModelPatch` — `createBackupUnique` retries with `-001`/`-002`/… counter suffix when the timestamped path already exists. Handles same-second double-patches and leftover backups. Tested.
- **#4 (revert clobbers hand-edits)**: Real fix — new `RevertConfirmModal` shows a current → backup unified diff before clobbering, so the user sees any edits they're about to lose. **DOM-probed end-to-end against OP-Test**: hand-edited the workflow post-patch, opened the dialog, clicked "Revert last patch", saw the diff (with the hand-added comment line marked `-`), confirmed → file restored from backup, .bak trashed. Big UX win.
- **#6 (interactive flag in URI handler)**: Real fix — `handleOpOpenAgentUri` now passes `interactive: false`. URI dispatchers can't participate in modal callbacks; an actionable Notice is the right surface.

**Waived (with rationale — not bugs):**
- **#5 (`launchModelOverride` recursion loop)**: The picker only offers entries from `MODEL_REGISTRY[chosenAgent]`, all of which validate via `validateModelName`. A user can hand-type something invalid, but `validateModelName` rejects it deterministically and the dialog re-opens — same UX as a fresh failure. No unbounded recursion: each iteration consumes a user input.
- **#7 (overlapping cross-agent aliases — `detectOverflowAgent` non-determinism)**: The current `MODEL_REGISTRY` has no overlapping aliases (alias collision would be a registry bug, not a recovery-dialog bug). If a future registry adds overlapping aliases, the canonical fix lives in `modelRegistry.ts`'s validator (lifting the conflict to a `validateRegistryShape` invariant), not here. Out of scope.
- **#8 (concurrent dialog instances)**: Not reproducible in normal Obsidian use — modals are stacked, not parallel; `app.workspace.openModal()` queues. The `onClose` cancellation guard is correct: a sibling dialog can't pre-empt this one. Would need a real multi-modal stack manager to break this, which Obsidian doesn't expose.
- **#9 (`vault.trash` system: true)**: System trash is the correct default. The `core-plugin: trash` "permanently delete" option is the user's explicit override of vault-side trash; `system: true` routes to the OS trash, which is recoverable on macOS/Windows even with that option enabled. Settings-controlling this would over-engineer for a fringe case.
- **#10 (`.bak-*` accumulation)**: Out of scope for v1 of OP-186 child 3e — the issue is "one-step undo only," not lifecycle management. A future ticket can add a `bakRetention: number` setting + a sweeper. Open as a follow-up.

**Final state**: vitest **1389 / 1389** post-fix (was 1385 pre-fix; +4 tests from Copilot). Live OP-Test DOM smoke re-run end-to-end after pulling: open advisory dialog → pick chip → patch → confirm diff → file written + `.bak-*` created → re-open dialog with hand-edited file → click Revert → confirm-with-diff modal renders the current→backup diff → restore → `.bak-*` trashed. All paths green.

Merging.

PR #268, opened 2026-04-26
OP-206: dry-run preview surface with auto-expand for first 3 launches (3f)

earchibald:
## Summary

OP-186 child 3f. Two entry points compose the FULLY-rendered launch prompt for a (workflow, mode, agent, issue) tuple — exactly the string the launcher would pass — both routed through `loadAndComposeWorkflow` so preview ≡ reality.

- **LaunchAgentModal preview disclosure** beneath the Launch button. Auto-expands on the first three launches per Obsidian session unless `previewAutoExpandDismissed` is set. Reactive to agent / launchVars changes (per-keystroke edits run the pure composer against the cached bundle — no vault re-read, no focus theft). Copy + dismiss actions.
- **Settings → Workflows "Preview composed prompt"** modal. Issue (sub-picker) / Mode (dropdown) / Agent (dropdown) toggles. Mirrors the launcher's lenient kickoff fallback so the preview matches reality even when the chosen mode's step isn't declared.
- **Pure auto-expand policy** in `previewAutoExpandPure.ts` (`shouldAutoExpand` — counter ≤ 3 ∧ ¬dismissed); session counter in `previewAutoExpand.ts` resets implicitly on Obsidian restart.
- **Settings flag** `previewAutoExpandDismissed: boolean` (default `false`).

Version: 0.71.0 → 0.72.0 (minor — additive).

## Test plan

- [x] `vitest` 1359/1359 (4 new pure cases on the auto-expand boundary)
- [x] `tsc --noEmit` — 8 pre-existing baseline errors, 0 new
- [x] Smoke test on OP-Test (DM-1 / demo project / `workflowMode: "modules"`):
  - [x] Launches 1/2/3: `aria-expanded="true"`, `<pre>` populated ("111 chars · 4 modules · 2 diagnostics"), Copy + Dismiss visible
  - [x] Launch 4: `aria-expanded="false"`
  - [x] Click Dismiss → `previewAutoExpandDismissed: true` persists; reload → next launch stays collapsed
  - [x] Settings "Preview composed prompt" modal renders 115-char prompt against DM-1's RenderContext; mode dropdown swap re-composes (kickoff fallback fires for `evaluate`)

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.72.0",
+  "version": "0.73.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.72.0",
+  "version": "0.73.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/explainWorkflow.ts
+++ b/plugins/op-obsidian/src/explainWorkflow.ts
@@ -146,6 +146,12 @@ function resolveProfileById(settings: OpSettings, raw: string): AgentProfile {
   return resolveProfile(settings, id);
 }
 
+// OP-206 (3f): `buildIssueRenderContext` and `readProjectVars` are exported
+// for reuse by `previewWorkflowModal.ts` so the Settings preview and the
+// launcher both build the same RenderContext. If you add a new field to either
+// function's output, check `previewWorkflowModal.ts` (`recompose()`) and
+// `explainWorkflow.ts` (`explainWorkflow()`) in lockstep — they share the
+// same interface contract and divergence produces a preview-vs-reality gap.
 export function buildIssueRenderContext(
   app: App,
   settings: OpSettings,
--- a/plugins/op-obsidian/src/explainWorkflow.ts
+++ b/plugins/op-obsidian/src/explainWorkflow.ts
@@ -146,7 +146,7 @@ function resolveProfileById(settings: OpSettings, raw: string): AgentProfile {
   return resolveProfile(settings, id);
 }
 
-function buildIssueRenderContext(
+export function buildIssueRenderContext(
   app: App,
   settings: OpSettings,
   entry: IssueEntry,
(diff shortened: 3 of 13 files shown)


earchibald:
@copilot please pressure-test this dry-run preview surface. Specific concerns:

1. **Auto-expand counter races.** The session counter is module-scoped (`previewAutoExpand.ts`) and bumps in the `LaunchAgentModal` constructor, *before* the modal opens. If the user cancels the launch picker upstream of `LaunchAgentModal` (e.g. `pickIssueInteractive` cancelled), does the counter still increment? What if two `op-open-agent-pick` modals are constructed back-to-back without the first opening — does the user lose two of their three "free" auto-expands?

2. **Per-keystroke compose cost.** `recomposeLightly` runs `composeWorkflowPure` on every keystroke in the launchVars input. For a workflow with many modules + a long composed text, is the per-keystroke cost noticeable? Is there a debounce I should add, or is the pure composer fast enough?

3. **`bundle.workflow` null path in `previewWorkflowModal.ts`.** When the project has no `WORKFLOW.md`, `bundle.workflow` is null and the kickoff fallback never fires. The `(no composed prompt — workflow file or modules missing)` hint shows. Is the user-facing copy clear that the issue is missing config, not a bug? Are there workflow files that load but produce empty composed text where this hint would be misleading?

4. **Dismiss is one-way.** Once `previewAutoExpandDismissed: true`, there's no UI to flip it back. Is that intentional friction or a regression risk for new users who dismiss accidentally? The Settings → Workflows "Preview composed prompt" entry doesn't surface a "re-enable auto-expand" toggle.

5. **`buildIssueRenderContext` / `readProjectVars` exported from `explainWorkflow.ts`.** I exported these for `previewWorkflowModal` to reuse. Does that risk drift between `op-explain-workflow` (CLI) and the preview modal if either path adds a new field?

6. **Race with parallel agents on OP-Test.** During smoke-testing, a parallel agent's dev-sync clobbered my synced bundle mid-test (timestamps in the README of OP-Test main.js). Is there a defensive check `dev-sync.mjs` should add (e.g. write-and-verify) so agents can detect mid-test interference? Out of scope for this PR but worth flagging.

7. **`<pre>` block whitespace round-trip.** I set `white-space: pre-wrap; word-break: break-word;` so long composed lines wrap. Are there cases (CR-only line endings, trailing spaces in module bodies, embedded `\r\n` from Windows-authored modules) where this CSS breaks fidelity vs. what the launcher actually sends?

8. **Modal teardown when Obsidian theme changes.** `op-launch-modal__preview-text` background uses `var(--background-secondary)`. Does the modal repaint correctly if the theme switches mid-session, or does the cached `<pre>` retain the old colors?

9. **Multi-issue picker performance.** `IssuePickerModal` consumes `this.plugin.store.issues()` — every issue across every project including resolved ones (we filter by `!e.resolvedFolder`). Is the picker noticeably slow with 1000+ issues? Should this be paginated or scoped to the active project?

10. **`navigator.clipboard.writeText` permission denial.** The Copy fallback uses `document.execCommand("copy")` if the modern API rejects. On Obsidian desktop (Electron), is the modern API always permitted, or do we ever hit the fallback path? Does the fallback trigger spurious focus changes that could close the modal?

PR #269, opened 2026-04-26
OP-190: OP-Test seed/workflow-modules + smoke harness for workflow modules

earchibald:
## Summary

Adds `seed/workflow-modules` to the OP-Test seed ladder and ships a re-runnable smoke harness for the OP-181 workflow-module pipeline. Other OP-181 children (OP-189 / OP-191 / OP-208 etc.) can chain `node scripts/smoke-workflow-modules.mjs` after their dev-sync.

- `scripts/build-seeds.mjs` — fifth seed cell. Lays down a global module (`Projects/_op-modules/branching.md`, kickoff scope, `reviewer_handle=@module-default` Module-default), a per-project module (`Projects/testing/MODULES/plan-rules.md`, plan scope), a per-project workflow file (`Projects/testing/WORKFLOW.md`, kickoff→[branching], plan→[plan-rules]), Project-default `vars:` on STATUS.md (`reviewer_handle: "@op-test-project"`), and launchable test issue `TST-5`. New helpers: `writeVaultFile`, `setProjectVars`, `setIssueAgent`.
- `scripts/reset-test-vault.mjs` — accepts `workflow-modules`.
- `scripts/smoke-workflow-modules.mjs` — invokes `op-explain-workflow` for kickoff + plan modes plus `op-list-vars`, then asserts composed-prompt content, Project-default precedence override, chunk shape per step, registry-context resolution, and zero error-severity diagnostics.
- `CLAUDE.md` — documents seed + harness under "Resetting between scenarios".

Plugin code untouched — meta-only repo change.

## Test plan

- [x] `node --check` on all three scripts (`build-seeds.mjs`, `reset-test-vault.mjs`, `smoke-workflow-modules.mjs`)
- [x] eval-body JS strings emitted by `writeVaultFile` / `setProjectVars` / `setIssueAgent` parse cleanly via `new Function(…)`
- [x] schema cross-check against `workflowFilePure.ts` / `workflowModulePure.ts` / `composeWorkflowPure.ts` for required frontmatter shape (`type`, `id`, `title`, `scope`, `schema: 1`, `default_agent`, `default_model`, `steps[].step`, `steps[].modules`)
- [x] smoke-harness assertion logic dry-run against a synthetic `ExplainWorkflowPayload` (OP-203 schema)
- [ ] **End-to-end deferred**: OP-Test on the test machine is currently mutated by the live OP-207 worker (uncommitted `Projects/demo/` from its editor-save-validator smoke test). The OP-Test git repo also has zero commits / no seed tags on this machine — needs a one-time initial commit before `build-seeds.mjs` can run end-to-end. Verification recipe captured in the issue note's Notes section: run `build-seeds → reset-test-vault workflow-modules → smoke-workflow-modules → build-seeds again` and assert `ALL PASS` + identical state.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -69,10 +69,20 @@ The OP-Test vault is a git repo with named seed tags capturing known states. Bef
 
 ```bash
 node scripts/reset-test-vault.mjs <seed>
-# valid seeds: empty | scaffolded | mid-flow | github-linked | multi-project
+# valid seeds: empty | scaffolded | mid-flow | github-linked | multi-project | workflow-modules
 ```
 
 The script asserts the OP-Test vault is open in Obsidian (focus not required since OP-175), runs `git reset --hard seed/<name> && git clean -fd` inside OP-Test, then reloads op-obsidian so Obsidian re-reads the vault state. The seed ladder itself is built (and rebuilt) by `node scripts/build-seeds.mjs` — re-runnable so seeds stay in sync as the plugin's scaffolding behavior evolves.
+
+### `seed/workflow-modules` + smoke harness
+
+`seed/workflow-modules` is the launchable checkpoint for the OP-181 workflow-module pipeline (one global module, one per-project module, a per-project workflow file referencing them, project-level `vars:`, plus issue `TST-5`). After resetting to it, run:
+
+```bash
+node scripts/smoke-workflow-modules.mjs
+```
+
+The harness invokes `op-explain-workflow id=TST-5 mode=kickoff` and `mode=plan`, plus `op-list-vars project=testing issue=TST-5`, then asserts the composed prompt contains the expected module bodies, that variables resolve through the precedence chain (project-default overrides the module's `name=VALUE` default), and that no `error`-severity diagnostics are emitted. Other workers landing changes in the OP-181 child series should run this against `seed/workflow-modules` after their dev-sync as part of the per-PR smoke step.
 
 ## Agent-Vault is BRAT-only
 
--- a/scripts/build-seeds.mjs
+++ b/scripts/build-seeds.mjs
@@ -229,10 +229,21 @@ build("seed/workflow-modules", () => {
     fail(
       `seed/workflow-modules: expected op-new to return TST-5 (cumulative ladder includes TST-1..4 from seed/mid-flow), got ${tst5.issueId}. ` +
         "If a new seed has been inserted between seed/mid-flow and seed/workflow-modules that adds testing-project issues, " +
-        "the smoke harness's hardcoded ISSUE = 'TST-5' must move in lockstep.",
+        "update the smoke-config.json write below and re-run.",
     );
   }
   setIssueAgent(tst5.path, "claude");
+
+  // Write a machine-readable config so the smoke harness doesn't hardcode the
+  // issue ID. The file is committed into the seed, so `reset-test-vault
+  // workflow-modules` always lands this exact JSON on disk before the harness
+  // reads it. If a future seed reshuffles issue numbering the build guard above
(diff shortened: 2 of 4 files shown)


earchibald:
@copilot please pressure-test this seed-ladder + smoke-harness change. Specific concerns to probe:

1. **Idempotency of the seed cell.** `build("seed/workflow-modules", …)` runs after `seed/multi-project`'s `op-new` already created TST-1..4. On a re-run, `ensureBaselineTag()` resets to `seed/empty` and replays all seeds; does the cumulative ladder still produce TST-5 deterministically? What happens if a future seed inserted between `multi-project` and `workflow-modules` adds another testing-project issue — does TST-5 silently shift to TST-6 and break the smoke harness? Should the harness derive the issue id from the build phase instead of hardcoding `TST-5`?

2. **MetadataCache race.** `writeVaultFile` calls `app.vault.create(path, content)` and returns immediately. The downstream `op-edit-module` / `op-explain-workflow` reads frontmatter via `metadataCache.getFileCache(file)?.frontmatter`. Is that cache populated synchronously after `vault.create`, or does the loader need a tick to settle? If there's a race, could `seed/workflow-modules` commit a state where the workflow file is on disk but the metadataCache hasn't yet parsed its frontmatter, leaving the next reset+launch in a half-loaded state?

3. **Project-default precedence assertion.** The smoke harness asserts `composed.text` contains `@op-test-project` AND does NOT contain `@module-default`. Does the composer ever emit the module-default value as a *fallback annotation* (e.g. inside a diagnostic block, as a "value would be" note in `perVarSourceMap.source`)? If so, the negative assertion is brittle — `text.includes("@module-default")` could fire on output that's actually correct. Confirm `composed.text` is body-only (no diagnostic prose).

4. **`op-list-vars` for a resolved-issue.** The harness asserts `id` row resolves to TST-5 *via* `op-list-vars project=testing issue=TST-5`. What if `entry.project` for TST-5 doesn't equal `"testing"` (e.g. if the future plugin shape uses a different slug field)? The handler throws — does that surface as a non-zero exit on the CLI or a silent JSON `error`? Verify the harness fails noisily on that path rather than passing-by-default.

5. **`scope: kickoff` on the global module vs `step: kickoff` on the workflow.** The composer uses `step.modules` lookup by id (not by `module.scope`). My module's `scope: kickoff` is metadata only. Is there any consumer (Settings UI, validator, dry-run) that *does* enforce `module.scope === step.step` and would treat my `branching.md` as misfiled if the workflow uses `step: kickoff`? If yes, the seed would emit a warning that breaks the "zero error-severity" assertion downstream.

6. **`op-new` filename collision.** `setIssueAgent` references `Projects/testing/ISSUES/TST-5 Workflow modules smoke target.md` by literal path. The plugin's filename sanitizer handles unicode + special chars; if anyone tweaks the title later, the literal path breaks. Worth deriving the path from `dispatch("op-new", …)`'s response payload (it returns `path`), like `setRelated` does, instead of hardcoding the filename.

7. **`processFrontMatter` on STATUS.md after `op-scaffold`.** `op-scaffold` writes the STATUS.md frontmatter; my `setProjectVars` then merges a `vars:` map. Does Obsidian's `processFrontMatter` preserve the existing `prefix:` / `repo_path:` fields, or does it round-trip-rewrite them and risk reordering / quoting drift?

8. **Reset-then-reload timing.** `reset-test-vault.mjs workflow-modules` does `git reset --hard seed/workflow-modules && git clean -fd`, then `obsidian plugin:reload id=op-obsidian`. If a workflow-module file was added on disk by the reset but op-obsidian's metadataCache hadn't yet re-scanned, the next `op-explain-workflow` could return `unknown-module`. Does the plugin reload force a full re-scan, or does it only reload the plugin's own state?

Thanks!

PR #270, opened 2026-04-26
OP-207: editor-save workflow validator (3g) — squiggles + status footer

earchibald:
## Summary

- Editor-save validator paints `WorkflowDiagnostic[]` from `composeWorkflow` as red/yellow CM6 squiggles inline + a status footer line under the editor for every save of a module / workflow file.
- Pure layer (`enumerateTuples`, `locateDiagnostic`, `dedupeDiagnostics`, `summarizeStatus`) sweeps every (step × installed agent) tuple, dedupes across the sweep, and maps diagnostics back to text ranges for the saved file.
- Bumped `op-obsidian` to `0.74.0` (minor — additive new editor surface).

## Files

Created:
- `plugins/op-obsidian/src/editorWorkflowValidatorPure.ts` (+`.test.ts`, 18 tests)
- `plugins/op-obsidian/src/editorWorkflowValidator.ts` (+`.test.ts`, 11 tests)
- `plugins/op-obsidian/src/editorWorkflowValidatorExtension.ts`

Edited:
- `plugins/op-obsidian/src/main.ts` — `registerEditorExtension(workflowValidatorExtension(deps))`
- `plugins/op-obsidian/styles.css` — `.op-validator-mark--{error,warning,info}` (wavy underline) + `.op-validator-footer`

## Behavior

- Squiggles: red for errors, yellow for warnings, accent-colored for info. Tooltip carries the formatted `diagnosticToLine` rendering.
- Footer: `N errors will block launch · 1W · 1I` shape when errors present, `Workflow OK` on a clean run, `1 warning · 1 notice` for non-blocking states.
- Re-validates on `vault.modify`, `metadataCache.changed` (when the changed file is the bound editor's), and any doc edit (debounced 250ms).
- Diagnostics with no localizable anchor (size-budget, schema-mismatch, import-collision across foreign modules) flow into the footer's hover tooltip.

## Test plan

- [x] `vitest` — 1422/1422 pass (29 new).
- [x] Smoke (vault=OP-Test): module file with `{{vars.bare_var}}` referenced but never declared → yellow squiggle on the token + `1 warning` footer.
- [x] Smoke (vault=OP-Test, source mode): `default_model: opuss` typo in `Projects/demo/WORKFLOW.md` → red squiggle on `opuss` + `1 error will block launch · 1W` footer. (Live-preview folds frontmatter, so squiggles inside the YAML fence are only visible in source mode — render-mode constraint, not a validator bug.)
- [x] Build green, dev-sync clean, plugin reload via `obsidian plugin:reload`.


earchibald:
@copilot — adversarial review please. Pressure tests below; please probe specifically these failure modes:

**Locator correctness**
- The word-boundary regex `[^A-Za-z0-9_.-]` (in `editorWorkflowValidatorPure.ts` `indexOfWord`) — does it correctly *fail* to match identifier-substring cases like `opuss-pro` when looking for `opuss`? Confirm the boundary tracks the registry's identifier alphabet.
- For a `bad-model` diagnostic where `extra.badName` happens to equal a substring of the agent id (e.g. `claude` vs an agent prose mention) — does the locator land on the model spec line or the wrong place?
- For `intra-scope-collision` when both implicated modules share the same basename across global + per-project scopes — does the basename-only check let one collide-implicated module's squiggle leak onto the other? (`Projects/_op-modules/foo.md` vs `Projects/demo/MODULES/foo.md` both basename `foo`.)
- `locateMissingVar` falls through from `{{vars.<name>}}` to `{{<name>}}`. If a body contains BOTH `{{vars.id}}` (user var, undeclared) and `{{id}}` (plugin var) — does the right token get squiggled?

**Race / lifecycle**
- ViewPlugin attaches `vault.on('modify')` + `metadataCache.on('changed')` + `workspace.on('active-leaf-change')` listeners directly on `app.*` rather than via `plugin.registerEvent`. They're cleaned up in `destroy()` — verify there's no scenario where the plugin reloads and orphans listeners (re-registration adds duplicates).
- Debounce: 250ms timer per ViewPlugin, cancelled on next schedule. If the editor closes mid-debounce, does `runValidate` early-out via `this.destroyed`? Check that `applyResult` can't re-dispatch onto a torn-down view.
- `currentRaw` is snapshotted at the top of `runValidate` from `view.state.doc.toString()` while `validateFile` is awaited (it does vault reads). If the user edits during the await, the locator could run against stale text. Is the resulting mismatch acceptable, or should we re-snapshot post-await?

**False positives at edit-time**
- `buildSyntheticContext` fills every plugin var with a sentinel string so `{{id}}` etc. don't fire missing-var. Does this hide *real* unknown plugin-var tokens (`{{badname}}`)? Trace through `renderTemplate` to confirm only registered names get the sentinel.
- Project vars from STATUS.md are deliberately NOT read at edit time. What does this miss — are there any STATUS-only vars whose absence would silently hide a real authoring bug?
- The validator falls back to the workflow's declared agents when the installed-agent intersection is empty (e.g. user has only `claude` installed, workflow declares `gemini`). Is that the right behavior? Or should the validator skip and report "no installed agent matches"?

**Render mode**
- Squiggles inside the YAML frontmatter only render in true source mode (Live Preview folds the fence). The footer line still surfaces in either mode. Should the footer call out "switch to Source mode to see the squiggle in the YAML"? Or is the footer's `1 error will block launch` count enough?

Please push fixes if you find clear-cut bugs; for judgment calls, comment with your analysis.

PR #273, opened 2026-04-26
OP-213: Reference docs auto-generated from registries + CI guard (10c)

earchibald:
## Summary

- Adds `docs/workflow-modules/reference/` with six pages — README index, plugin-vars, module-schema, workflow-schema, precedence, scope-tags. Plugin-vars and workflow-schema carry auto-generated tables sourced from `pluginVarRegistry.ts` (every always-on var + the resolved `PARENT_NONE_SENTINEL`) and `modelRegistry.ts` (per-agent aliases + canonical versioned ids). Hand-written pages defer to `docs/specs/workflow-{module,file}-schema.md` for the canonical field-by-field contracts.
- Adds `scripts/check-workflow-docs.mjs` — dependency-free Node script that reads the two TypeScript registries as text, pulls the literal initializer entries via narrow regex + a balanced-brace walker, and rewrites the auto-generated blocks in place between `<!-- AUTO-GENERATED:<key> -->...<!-- /AUTO-GENERATED:<key> -->` sentinels. `--check` regenerates and `git diff --exit-code`s the affected paths so drift fails CI.
- Adds a `workflow-docs-drift` job to `.github/workflows/op-obsidian-pr-checks.yml`. Always-runs shape with step-level `if:` gating on touching the registry sources, the script, the reference docs, or the workflow file itself.

Closes the OP-193 (Child 10) grandchild **10c** — see umbrella for sibling work (10a/OP-211 conceptual overview + quickstart already merged; 10b/OP-212 authoring tutorials in flight; 10d/OP-214 sharing/migration; 10e troubleshooting; 10f worked examples).

No `plugins/op-obsidian/` source edits — docs + script + CI only. No version bump.

## Test plan

- [x] `node scripts/check-workflow-docs.mjs` — first run wrote `plugin-vars.md`, second run unchanged. Idempotent.
- [x] `node scripts/check-workflow-docs.mjs --check` on clean tree — exit 0, "workflow-docs check: ok".
- [x] Drift simulation — copied registry to `/tmp`, inserted a synthetic `fake_drift_var`, pointed generator via `OP_DOCS_PLUGIN_VAR_REGISTRY`. `--check` exited 1 with a structured message naming the page + the recovery command + the diff. Restored clean afterward.
- [x] Hard-error path — renamed `PARENT_NONE_SENTINEL` in a fake registry; script exited 2 with a precise "could not find … update extractParentNoneSentinel()" pointer.
- [x] Hard-error path — stripped sentinels from a managed page; script exited 2 with "no AUTO-GENERATED sentinels found …".
- [ ] CI: confirm the new `workflow-docs-drift` job runs (and is gated correctly — should run on this PR since it adds the docs and the script, but not on PRs that don't touch the registries/docs/script).

## @copilot adversarial review

Pressure test:

1. **Sentinel-marker robustness.** Could a hand-written page with a malformed sentinel (extra whitespace, wrong-cased key, missing closing marker) produce a corrupted file that *still passes* `--check`? E.g. a single open marker with no close — does the regex skip it silently or fail loudly? What about nested markers (`AUTO-GENERATED:a` containing another `AUTO-GENERATED:b`)? Should they be allowed?
2. **Regex extraction fragility.** `extractPluginVars`'s entry regex requires keys in the exact order `name, description, example, compute`. Is a key reorder in `pluginVarRegistry.ts` a silent miss (entries not found, table empty) or a loud failure (the `entries.length === 0` check fires)? Same for `extractModelRegistry`'s `aliases:` then `versioned:` ordering. If silent: harden it.
3. **Template-literal escape coverage.** `renderStringOrTemplate` decodes `\``, `\\`, `\$`, `\n`, `\t`, `\r`. What if a future registry entry uses `é` or `\xff`? The "unknown escape" branch preserves the literal `\u` — does that produce a broken table cell? Should I either decode them or fail loudly instead of silently passing through?
4. **Markdown cell escaping.** `escapeCell` only escapes `|` and `\n`. A description containing literal `<` or HTML-like content would render as raw HTML in some markdown processors. Acceptable? If the registry adds a `description: "use <foo> here"` entry, would that produce surprising output?
5. **`git diff --exit-code` behavior in `--check`.** What if the working tree has *unstaged* edits to the reference docs that aren't drift (e.g. a hand-written prose change in the same PR)? The script regenerates first then diffs, so any working-tree drift — generated or hand-edited — fails the check. Is that the right semantic? Could it produce false-positive CI failures on PRs that legitimately edit prose around the auto-gen block?
6. **CI gate path-filter completeness.** The `workflow-docs-drift` step-level `if:` gates on the four file patterns. Is there any path where the registry could change *without* one of those files appearing in the diff (e.g. via a refactor that moves `MODEL_REGISTRY` to a new file)? The CI should follow.
7. **Drift reset symmetry.** If a CI run regenerates files in the runner's checkout, the runner's working tree is now dirty — but CI doesn't push back. Confirm that `--check` is purely *read*ing the diff (no commit, no push) and that the runner's mutation is harmless.
8. **Backslash-escape interpretation.** Verify that the `parent` var's description `\`(none — …)\`` renders as literal backticks `\`(none — …)\`` in the generated table (forming an inline code span on render), not as the two characters `\\\``. The fix in this PR walks the template literal char-by-char to decode escapes; sanity-check that no other entry in the registry trips this path.

### DIFF
--- a/.github/workflows/op-obsidian-pr-checks.yml
+++ b/.github/workflows/op-obsidian-pr-checks.yml
@@ -46,3 +46,41 @@ jobs:
         if: steps.op_changes.outputs.relevant == 'true'
         working-directory: plugins/op-obsidian
         run: npm ci --dry-run
+
+  workflow-docs-drift:
+    name: Workflow docs drift (auto-generated reference tables)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # Always-run job with step-level if: gating, same shape as the
+      # lockfile-sync job above so the check shows up cleanly in
+      # required-status UX without blocking unrelated PRs.
+      - name: Detect workflow-docs-relevant changes
+        id: docs_changes
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          if git diff --name-only "$BASE_SHA" "$HEAD_SHA" \
+               | grep -qE '^plugins/op-obsidian/src/(pluginVarRegistry|modelRegistry)\.ts$|^scripts/check-workflow-docs\.mjs$|^docs/workflow-modules/reference/|^\.github/workflows/op-obsidian-pr-checks\.yml$'; then
+            echo "relevant=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "relevant=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Setup Node
+        if: steps.docs_changes.outputs.relevant == 'true'
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+
+      # Regenerates the auto-generated tables in docs/workflow-modules/reference/
+      # from the registry sources, then `git diff --exit-code`s the affected
+      # paths. Fails CI when the registry was edited without running the
+      # generator — prevents the "single source of truth" claim from drifting.
+      - name: check-workflow-docs --check
+        if: steps.docs_changes.outputs.relevant == 'true'
+        run: node scripts/check-workflow-docs.mjs --check
--- a/docs/workflow-modules/reference/README.md
+++ b/docs/workflow-modules/reference/README.md
@@ -1,0 +1,72 @@
+# Workflow Modules — Reference
+
+This folder holds the canonical reference for the workflow-modules
+subsystem. Three categories of page:
+
+- **Auto-generated tables** that ride alongside hand-written prose. The
+  tables come from a registry source in `plugins/op-obsidian/src/` and
+  are kept in sync by `scripts/check-workflow-docs.mjs`. Editing the
+  table by hand will be undone the next time the script runs.
+- **Hand-written cheat-sheets** that point at the canonical specs in
+  `docs/specs/` for the deep field-by-field contract.
+- **Hand-written precedence and tag references** that explain how the
+  pieces compose at runtime.
+
+## Pages
+
+- [`plugin-vars.md`](./plugin-vars.md) — every always-on plugin variable
+  the renderer surfaces (`{{id}}`, `{{title}}`, `{{branch}}`, …),
+  generated from `pluginVarRegistry.ts`.
+- [`module-schema.md`](./module-schema.md) — workflow-module frontmatter
+  cheat-sheet and diagnostic codes. Points at
+  [`workflow-module-schema.md`](../../specs/workflow-module-schema.md)
+  for the canonical contract.
+- [`workflow-schema.md`](./workflow-schema.md) — workflow-file frontmatter
+  cheat-sheet, plus the auto-generated **model registry** of allowed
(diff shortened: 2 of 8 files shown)


earchibald:
@copilot please review this PR. Specific adversarial pressure tests I want answered (numbered to make it easy to address each):

1. **Sentinel-marker robustness.** A page with an unmatched marker (open `<!-- AUTO-GENERATED:foo -->` with no closing `<!-- /AUTO-GENERATED:foo -->`) — does the script's regex skip it silently or fail loudly? What about nested same-key markers, or two `AUTO-GENERATED:foo` blocks on one page? Verify the failure modes match the comments in `scripts/check-workflow-docs.mjs:rewriteOnePage`.

2. **Regex extraction key-order fragility.** `extractPluginVars` requires keys in the order `name, description, example, compute`. If a future PR reorders `pluginVarRegistry.ts` entries to e.g. `description, name, example, compute`, does the regex silently produce zero entries (caught by the `entries.length === 0` guard) or silently produce a wrong table? Same question for `extractModelRegistry` (`aliases:` then `versioned:` order).

3. **Template-literal escape coverage.** `renderStringOrTemplate` decodes `\``, `\\`, `\$`, `\n`, `\t`, `\r`. What happens for `é`, `\xff`, `\0`? The "unknown escape" branch preserves the literal `\u…` — could that produce broken or surprising markdown? Should the script fail loudly on unknown escapes instead of preserving them?

4. **Markdown cell escaping.** `escapeCell` only escapes `|` and `\n`. A description containing `<foo>` would render as raw HTML in some processors. A description containing a literal markdown table separator `---` mid-cell? Acceptable today, or do I need to widen escaping?

5. **`git diff --exit-code` semantics in `--check`.** The script regenerates first, then diffs. If a PR also legitimately edits the surrounding hand-written prose in the same reference page, the diff will include those prose edits — does that produce a false-positive CI failure? Walk through: PR edits `plugin-vars.md`'s prose; the auto-gen block is in sync with the registry; CI runs `--check`; what does the diff show?

6. **CI gate path-filter completeness.** The gate matches `plugins/op-obsidian/src/(pluginVarRegistry|modelRegistry)\.ts$`. If a refactor moves either registry to a new file or splits it, the gate skips and CI silently passes. Is the right mitigation to add a smoke-test that ensures the script can find both registries via `node scripts/check-workflow-docs.mjs --check` even when no PR-relevant files changed? Or to broaden the gate?

7. **Drift-reset symmetry.** On a CI runner, `--check` mutates the working tree (regenerates the files). The runner doesn't push back. Confirm the regeneration is idempotent and harmless on the runner (no side effects beyond the working-tree edits the runner discards).

8. **Backslash-escape regression.** The `parent` var's description in `pluginVarRegistry.ts:123` is a template literal with `\`...\``. The first run of the generator produced literal `\`` in the markdown table; the fix walks the template char-by-char to decode escapes. Confirm the rendered table cell on `plugin-vars.md:50` shows `(none — this is a top-level issue)` wrapped in backticks (forming an inline code span), not the two-character sequence `\``. And confirm no other registry entry trips this same path.

Thanks.

PR #274, opened 2026-04-26
OP-215: 10e — troubleshooting + FAQ + settings reference + in-product help links

earchibald:
## Summary

- Three new docs under `docs/workflow-modules/` (troubleshooting, FAQ, settings-tab reference). Troubleshooting has one section per shipped `WorkflowDiagnosticCode` (8 codes — the issue's three ghost codes don't exist in the plugin); settings-reference uses ASCII layouts per the OP-211 "no screenshots" precedent.
- In-product help links: a `?` icon on the Workflows H2, the Available variables collapsible header, and the Injection collapsible header opens the relevant doc anchor on GitHub. `FormattedDiagnostic` gains a `helpUrl` field so `diagnosticToBlock` auto-stamps a `More: <url>` footer in CLI / editor / preview surfaces.
- CI guard: `plugins/op-obsidian/src/docLinks.test.ts` parses every doc the registry references and fails on any missing anchor (github-slugger semantics — em-dash collapses to consecutive hyphens, no whitespace collapse).
- Bumped 0.74.1 → 0.74.2 (patch — additive doc + UI affordance, no behavior change).

## Test plan

- [x] `npx vitest run` (full suite): 1438 pass, 0 fail
- [x] OP-Test smoke: `app.setting.openTabById("op-obsidian")` shows 3 `.op-doc-link` icons with the right hrefs (`workflows`, `workflowsVars`, `injection` anchors).
- [x] OP-Test console clean after the settings panel render.
- [x] Existing `data-op-section` selectors still work (8 advanced collapsibles, 5 daily rows, Workflows H2 intact).

## Notes for review

`@copilot` — focus pressure points:
1. The github-slugger approximation in `docLinks.test.ts` — does it correctly model anchors with em-dashes (`Injection — what the migration changed` → `injection--what-the-migration-changed`, double hyphen)? Verify against any other em-dash heading.
2. `helpUrl` was added to `FormattedDiagnostic` and threaded through `diagnosticToBlock` only — not the one-line `diagnosticToLine`. Is the asymmetry justified, or should the one-line surface also stamp a footer?
3. `addDocLink(parentEl, sectionId)` opens the URL in a new tab via `target=_blank rel=noopener`. Is the `?` glyph + `aria-label` sufficient for accessibility?
4. The Injection collapsible's `?` icon is wired via `ADVANCED_SECTIONS_WITH_DOCS: Set<AdvancedSectionId>` rather than checking `sectionDocAnchor` for membership. Is that registry-vs-set duplication worth removing in favor of a try/catch on the lookup?
5. Scope reconciliation — the issue scope listed 3 diagnostic codes (`screenshot-permission`, `stalled`, `orchestrator-degraded`) that don't exist in `WorkflowDiagnosticCode`. The PR ships docs for the 8 real codes only. Sound?

### DIFF
--- a/docs/workflow-modules/03-troubleshooting.md
+++ b/docs/workflow-modules/03-troubleshooting.md
@@ -1,0 +1,275 @@
+# Workflow Modules — Troubleshooting
+
+Every error and warning the workflow-modules engine emits is a
+`WorkflowDiagnostic` — one shape, one formatter, every surface. This page
+walks each diagnostic code, explains what it means in plain English, and
+shows how to fix it.
+
+## Where the same diagnostic shows up
+
+A single diagnostic record (one `WorkflowDiagnostic`) appears in multiple
+surfaces — they all read it through the unified formatter
+(`workflowDiagnosticFormat.ts`), so the prose is byte-identical no matter
+where you encounter it:
+
+- **Settings → Workflows panel.** The module list shows per-module
+  diagnostics inline (badge + label + location + message + hint).
+- **`op-explain-workflow` / `op-list-vars` CLIs.** The structured payload
+  in `Projects/_scratch/op-last-response.md` carries `diagnosticLines`
+  (one-line) and `diagnosticBlocks` (multi-line); the stdout summary
+  reports per-severity counts.
+- **Editor squiggles.** Saving a `WORKFLOW.md` or module file runs the
+  loader and renders diagnostics as red/yellow underlines with the same
+  message and a status-footer count.
+- **Dry-run preview banner** (launch modal). The `▶ Composed prompt
+  preview` disclosure header shows the diagnostic count next to the
+  module count and char count.
+
+If a fix below says "edit the module" and you're staring at the Settings
+panel, you can open the file straight from the diagnostic row's location
+breadcrumb — clicking opens the file in the active leaf.
+
+## Severity legend
+
+- **Error** (`E`) — the workflow won't compose, or the module/file is
+  dropped from the load. Block-and-fix.
+- **Warning** (`W`) — composition continues, but something looks wrong
+  enough that the agent might do the wrong thing. Worth investigating.
+- **Info** (`I`) — advisory; the surface is showing you a guardrail or a
+  size notice. Not a failure.
+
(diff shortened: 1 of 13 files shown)


earchibald:
@copilot — adversarial review please. Concrete pressure tests:

1. **Slug correctness for em-dash headings.** The CI guard slugifier (`docLinks.test.ts`) treats `—` (em-dash) as a strip-then-spaces-survive pair. Heading `Injection — what the migration changed` should yield `injection--what-the-migration-changed` (double hyphen, no collapse). Verify against any *other* em-dash heading in the docs and confirm GitHub's anchor matches our test's expectation. If the approximation drifts on a real heading we don't have today (e.g. unicode letters, multiple consecutive em-dashes, en-dash `–`), the registry won't catch it.
2. **`helpUrl` asymmetry.** `diagnosticToBlock` now appends `More: <url>` but `diagnosticToLine` does not. Squiggle status footers and dry-run banner read both — does either render the line form somewhere where the missing footer is a regression vs. the new contract?
3. **`ADVANCED_SECTIONS_WITH_DOCS` duplication.** `docLinks.ts` already encodes which section ids have a target; `settings.ts` re-encodes that as a separate `Set`. Adding a new section means editing both files. Is `sectionDocAnchor` worth widening to return `undefined` so the call site can do an existence check instead?
4. **Settings drag-reorder regression.** The `?` icon was added to `varsCollapsible.header`, which is a `role=button` element. Does clicking the link bubble to the collapsible toggle? Test: with the panel open, click the `?` — does the collapsible close as a side effect?
5. **GitHub URL format.** The hard-coded prefix is `https://github.com/earchibald/obsidian-projects/blob/main/`. If the repo moves or a release is pinned to a tag, every help link 404s. Is the registry-of-one approach the right shape, or should this be a settings field?
6. **Scope drift call-out.** The issue scope listed `screenshot-permission`, `stalled`, `orchestrator-degraded` as `WorkflowDiagnostic` codes. They don't exist in the plugin. PR ships docs for the 8 real codes only. Confirm the omission is correct (i.e. these aren't aspirational codes due imminently).

PR #275, opened 2026-04-26
OP-216: 10f — worked-example library at docs/examples/workflow-library/

earchibald:
## Summary

Ships the **worked-example library** for OP-193's Child 10 (user docs umbrella) — a curated set of reusable workflow modules + reference workflows under `docs/examples/workflow-library/`, distilled from this project's `WORKFLOW.md` with per-project specifics swapped out for `vars:` defaults so any project can copy the modules in verbatim.

**8 modules** (within the 6–10 target):

| File | Scope | Vars (with defaults) | Codifies |
| :--- | :--- | :--- | :--- |
| `orient.md` | `kickoff` | _none_ | Read the issue note, surface ambiguity. |
| `branching.md` | `kickoff` | `default_branch=main`, `branch_strategy=worktree` | Branch / worktree discipline. |
| `tmux-safety.md` | `kickoff` | `protected_session_prefixes`, `experiment_session_pattern` | Don't kill shared tmux sessions. |
| `commit-mapping.md` | `implement` | `commit_command`, `commit_field` | Per-commit append to issue's commits list. |
| `pr-required.md` | `review` | `default_branch`, `pr_title_pattern`, `merge_command` | PR-required gate; squash-merge cadence. |
| `adversarial-review.md` | `review` | `reviewer_handle=@copilot`, `bypass_criteria=all` | Adversarial review with concrete pressure-test prompts; bypass criteria. |
| `version-cadence.md` | `finalize` | `package_name`, `bump_command` | Lockstep version bump at resolve time. |
| `gh-issue-close.md` | `finalize` | `auto_close_setting_path` | Don't close GH issue manually — let `op-resolve` handle it. |

Every module declares its `vars:` in one of the three documented shapes (bare / shorthand / object); together they exercise all three at least once, so the collection doubles as a vars-syntax reference.

**3 reference workflows**:

- `Projects/_op-workflow.md` — global default. Five steps (`evaluate → plan → implement → review → finalize`) — the historical OP-188 sequence.
- `Projects/code-project/WORKFLOW.md` — `extends:` the global default with `steps: []` (the canonical "inherit verbatim" shape).
- `Projects/docs-project/WORKFLOW.md` — standalone domain variant. Three steps; explicitly omits `review`/`finalize` because the format has no "remove inherited step" semantics.

## CI

`plugins/op-obsidian/src/docsExamples.test.ts` grows two `ExampleSpec` entries for the library (one per project slug). The manifest-parity test now dedupes so multiple slugs in one subtree are tolerated. The library's `plan` step ships intentionally empty (slot for projects to fill in their own planning checklist), so the test composes at every step *except* `plan`.

Full vitest run: 1438 pass / 0 fail.

## Version

op-obsidian 0.74.1 → 0.74.2 (patch — additive docs + test, no runtime change).

## Test plan

- [x] `npx vitest run` — 1438 pass / 0 fail
- [x] `node scripts/bump-version.mjs patch` — version files in lockstep, main.js fresher than manifest
- [x] `node scripts/dev-sync.mjs` — plugin reloaded into OP-Test
- [x] `obsidian vault=OP-Test eval code='app.plugins.plugins["op-obsidian"]?.manifest?.version'` — returns `0.74.2`
- [x] OP-Test dev console clean after reload (no errors logged on plugin load)

### DIFF
--- a/docs/examples/README.md
+++ b/docs/examples/README.md
@@ -8,11 +8,12 @@ ship to CI to keep the docs from drifting away from the engine.
 
 ## What's here
 
-| Subtree | Tutorial it backs | What it shows |
+| Subtree | Backs | What it shows |
 | :--- | :--- | :--- |
-| [`author-first-module/`](./author-first-module/) | [03 — Author your first module](../workflow-modules/03-author-your-first-module.md) | A single global module, no vars, no workflow file. The minimum viable file. |
-| [`compose-first-workflow/`](./compose-first-workflow/) | [04 — Compose your first workflow](../workflow-modules/04-compose-your-first-workflow.md) | Global default workflow + a per-project workflow that `extends:` it, with two modules and a per-step agent override. |
-| [`variables-and-templating/`](./variables-and-templating/) | [05 — Use variables and templating](../workflow-modules/05-variables-and-templating.md) | Plugin vars (`{{branch}}`, `{{today}}`, `{{repo_path}}`), user vars (`{{vars.package_name}}`), `name=VALUE` inline defaults, and the precedence chain in action. |
+| [`author-first-module/`](./author-first-module/) | Tutorial: [03 — Author your first module](../workflow-modules/03-author-your-first-module.md) | A single global module, no vars, no workflow file. The minimum viable file. |
+| [`compose-first-workflow/`](./compose-first-workflow/) | Tutorial: [04 — Compose your first workflow](../workflow-modules/04-compose-your-first-workflow.md) | Global default workflow + a per-project workflow that `extends:` it, with two modules and a per-step agent override. |
+| [`variables-and-templating/`](./variables-and-templating/) | Tutorial: [05 — Use variables and templating](../workflow-modules/05-variables-and-templating.md) | Plugin vars (`{{branch}}`, `{{today}}`, `{{repo_path}}`), user vars (`{{vars.package_name}}`), `name=VALUE` inline defaults, and the precedence chain in action. |
+| [`workflow-library/`](./workflow-library/) | Curated reusable library | Eight `vars:`-driven modules distilled from this project's `WORKFLOW.md` (branching, tmux safety, PR-required, version cadence, GH issue close, adversarial review, commit mapping, orient) plus three reference workflows (global default + extends-default + standalone docs-only). Copy into your own vault to bootstrap a project. |
 
 ## How the layouts map onto your vault
 
--- a/docs/examples/workflow-library/Projects/_op-modules/adversarial-review.md
+++ b/docs/examples/workflow-library/Projects/_op-modules/adversarial-review.md
@@ -1,0 +1,51 @@
+---
+id: adversarial-review
+title: Adversarial AI review before merge
+type: workflow-module
+scope: review
+order: 20
+vars:
+  - reviewer_handle=@copilot
+  - { name: bypass_criteria, default: "all", description: "How many bypass criteria must apply to skip review — 'all' (every criterion) or 'any' (any single one). 'all' is conservative; 'any' is unsafe." }
+---
+
+Every PR gets an adversarial review from {{vars.reviewer_handle}} (or
+equivalent reviewer agent) **before** merge, unless the diff meets the
+bypass criteria below.
+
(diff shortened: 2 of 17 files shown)


earchibald:
@copilot — please pressure-test this PR. Concrete prompts:

1. **YAML coercion of shorthand `vars:` defaults.** The `yaml` package parses the frontmatter; do any of the shorthand values get coerced to a non-string? Specific ones to check:
   - `experiment_session_pattern=op-experiment-\$\$` — does `\$\$` survive verbatim?
   - `reviewer_handle=@copilot` — `@` is a YAML reserved indicator at the start of a plain scalar; it's mid-string here, but worth confirming the `yaml` package doesn't reject it.
   - `bump_command=node scripts/bump-version.mjs` — multi-word value with `.` in `bump-version.mjs`; ensure it's parsed as one string, not a list.
   - `merge_command=gh pr merge --squash --delete-branch` — same concern.
   - `pr_title_pattern: \"<ID>: <subject>\"` (object form) — the `:` inside the value would have broken shorthand; confirm object form parses cleanly with the embedded colon.

2. **Empty `steps: []` under `extends:`.** `code-project/WORKFLOW.md` uses `extends: Projects/_op-workflow.md` + `steps: []`. The test passes, meaning the parent's full step list comes through. Is this the intended canonical \"inherit verbatim\" shape, or is the test passing by accident? Confirm against `workflowFile.ts` / `composeWorkflowPure.ts` merge logic. If the canonical shape is something else (e.g., omitting `steps:` entirely is supposed to fall back to legacy-3, which would break the PR's claim), the README and the file need rewording.

3. **Manifest-parity test dedupe.** I changed `inManifest = EXAMPLES.map((e) => e.name).sort()` to `Array.from(new Set(...)).sort()` so two manifest entries (`code-project`, `docs-project`) for the same `workflow-library` directory don't break parity. Does this loosen the safety check too far? E.g., could someone now sneak in a manifest entry whose name doesn't match any directory and the dedupe would hide it? (I think no — we still assert `onDisk == inManifest` after dedupe — but want a sanity check.)

4. **Intra-scope var collisions.** Within each scope:
   - `kickoff`: `branching` (`default_branch`, `branch_strategy`), `tmux-safety` (`protected_session_prefixes`, `experiment_session_pattern`), `orient` (none).
   - `review`: `pr-required` (`default_branch`, `pr_title_pattern`, `merge_command`), `adversarial-review` (`reviewer_handle`, `bypass_criteria`).
   - `finalize`: `version-cadence` (`package_name`, `bump_command`), `gh-issue-close` (`auto_close_setting_path`).
   - `implement`: `commit-mapping` only.

   Note `default_branch` appears in *both* `branching` (kickoff) and `pr-required` (review) — different scopes, so no `intra-scope-collision`. But check whether the shared name across scopes leads to surprising behavior at compose time when a user sets `vars.default_branch=develop` at the project layer (it should shadow both modules' defaults — same name, same intent — which I believe is correct).

5. **`bypass_criteria=all` wording.** The module body reads: \"`{{vars.bypass_criteria}}` must apply, or no bypass\". Renders cleanly as \"all must apply\" with the default. But if a user overrides to `any`, it reads \"any must apply\" — which is grammatically valid but flips the safety semantics (\"all\" → conservative; \"any\" → dangerous). Is this acceptable as a footgun, or should `bypass_criteria` instead be a fixed enum check at compose time? I lean toward leaving it — the module's docstring already says \"'all' is conservative; 'any' is unsafe\" — but want a second opinion.

6. **Cross-subtree var name reuse.** `package_name` appears in both this library's `version-cadence.md` and the existing `variables-and-templating/Projects/_op-modules/version-cadence.md`. Different example subtrees → different fake vaults → loaded separately by the test. README explicitly notes this. Verify there's no path through the loader where files from multiple `docs/examples/<subtree>/` directories end up in one fake vault simultaneously.

7. **Schema sanity.** `_op-workflow.md` declares `project: _global`. Is `_global` a legal slug, or does the loader reject underscore-prefixed slugs? (The existing `compose-first-workflow/Projects/_op-workflow.md` uses the same value, so I expect it's fine — but worth a check against `workflowFilePure.ts` validation.)

Generic \"please review\" yields a generic answer — please go through each numbered point above with concrete code references.

PR #276, opened 2026-04-26
OP-188: replace flowOrchestrator hardcoded matrix with workflow-file reader

earchibald:
## Summary

- Replace `flowOrchestrator.ts`'s hardcoded transition matrix with a walker over the project's loaded `WorkflowFile.steps`. Permanent `LEGACY_FLOW_ALIAS` map (`planning → plan`, `implementation → implement`, `finalization → finalize`) keeps existing in-flight issues' frontmatter parseable; new writes use canonical step ids only.
- Legacy fallback: when `workflow === null` or `workflow.source.isLegacy === true`, delegate to a private `legacyFlowAdvanceDecision` that runs the pre-OP-188 hardcoded matrix. Pre-modules projects keep auto-advancing until OP-189 migrates their WORKFLOW.md to the modules schema.
- `scaffoldProject.ts` seeds every new project with a default `WORKFLOW.md` shipping the canonical evaluate → plan → implement → review → finalize sequence.
- `setFlow.ts` `Flow` type becomes a free-form non-empty string — the workflow-file walker is now the semantic gate. `uriHandlers.ts` and `cliHandlers.ts` drop their hardcoded `FLOW_ENUM` rejection.
- `main.ts:advanceFlowAndLaunch` loads the project's workflow via `loadWorkflowFile` before invoking the orchestrator. Level-4 launch override carry-forward is unchanged.
- Bump op-obsidian to **0.75.0** (additive feature).

## Test plan

- [x] `npx vitest run` — 1466/1466 passing in op-obsidian (full suite, includes 30+ new flowOrchestrator tests covering canonical / legacy / custom workflows + legacy-fallback path).
- [x] `node scripts/bump-version.mjs minor` — manifest/package/plugin.json all at 0.75.0; build green.
- [x] `node scripts/dev-sync.mjs` to OP-Test; plugin reload confirmed.
- [x] `obsidian vault=OP-Test op-scaffold slug=op188-smoke prefix=SMK title=...` — verified `Projects/op188-smoke/WORKFLOW.md` lands with the canonical 5-step sequence and parses cleanly via `op-get-workflow`.
- [x] `op-set-flow flow=kickoff` (free-form id) and `flow=plan` (canonical) both accepted; whitespace-only rejected.

## Adversarial review pressure-test prompts (for @copilot)

1. **Alias map applied to both seams.** Verify `LEGACY_FLOW_ALIAS` is consulted both when reading the issue's `flow:` value AND when matching workflow file step ids. Confirm a `flow: planning` issue against a workflow file with `step: plan` advances correctly, and the reverse (`flow: plan` against `step: planning`) also walks.
2. **Evaluate-step skip semantics with non-canonical intervening steps.** `flowOrchestrator.test.ts:WITH_LINT` covers a workflow with `evaluate → lint-pass → plan → …`. The walker returns `null` (not `lint-pass`) when complexity is `simple` because `lint-pass` doesn't alias-resolve. Is that the right behavior, or should the walker scan further?
3. **Legacy fallback boundary.** When `workflow.source.isLegacy === true` but `steps:` happens to exist (synthesized kickoff), the fallback path runs the hardcoded matrix instead of the walker. Does this hide a real configuration error from the user? Should it emit a diagnostic?
4. **Race between `loadWorkflowFile` and `metadataCache`.** `advanceFlowAndLaunch` calls `loadWorkflowFile` immediately after `setFlow` writes the new flow value. `loadWorkflowFile` reads the workflow via `metadataCache`, which may not have flushed the post-`setFlow` rename if the issue file is being moved (it's not in this code path, but verify).
5. **Scaffold seed when project has zero step modules.** A workflow with all-empty `modules: []` step lists composes to an empty injection. Per OP-197 / OP-199, that's expected (no-op composition), but verify the module-injection chain doesn't error on an empty modules list — particularly the kickoff-fallback path in `composeWorkflowSection`.
6. **Loosened `setFlow` validation security/UX.** Now that `flow:` accepts any non-empty trimmed string, can a malicious or fat-fingered URI value land something problematic in frontmatter (e.g. `flow:\n  type: evil`)? `processFrontMatter` writes a string scalar so YAML injection is bounded — but verify `op-set-flow URI` doesn't accept embedded newlines.
7. **`nextFlow` field rename impact.** The `FlowAdvanceOutput` type adds `nextStep` and keeps `nextFlow` as an alias for wire-format compat. Verify no consumer reads `nextFlow` expecting the legacy enum names (e.g. `planning`); the field always carries canonical ids now.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.74.2",
+  "version": "0.75.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.75.0",
+  "version": "0.75.1",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.74.2",
+  "version": "0.75.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.75.0",
+  "version": "0.75.1",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/cliHandlers.test.ts
+++ b/plugins/op-obsidian/src/cliHandlers.test.ts
@@ -164,8 +164,16 @@ describe("parseSetFlowParams", () => {
     const r = parseSetFlowParams({ id: "OP-1", flow: "planning", complexity: "complex" });
     expect(r.ok && r.value).toEqual({ id: "OP-1", flow: "planning", complexity: "complex" });
   });
-  it("rejects invalid flow enum", () => {
-    const r = parseSetFlowParams({ id: "OP-1", flow: "wat" });
+  it("accepts free-form flow values (workflow walker enforces at advance time)", () => {
+    // OP-188: post-modules, parseSetFlowParams accepts any non-empty step id
+    // — the orchestrator's workflow-file walker decides what's valid for the
+    // project's actual workflow. Values not in the workflow simply produce
+    // a no-op auto-advance.
+    const r = parseSetFlowParams({ id: "OP-1", flow: "kickoff" });
+    expect(r.ok && r.value).toEqual({ id: "OP-1", flow: "kickoff" });
+  });
+  it("rejects whitespace-only flow value", () => {
+    const r = parseSetFlowParams({ id: "OP-1", flow: "   " });
     expect(r.ok).toBe(false);
     if (!r.ok) expect(r.error).toMatch(/invalid --flow/);
   });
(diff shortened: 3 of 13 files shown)


earchibald:

1. **Alias map applied to both seams.** `LEGACY_FLOW_ALIAS` (planning↔plan, implementation↔implement, finalization↔finalize) is consulted both when reading the issue's `flow:` value AND when matching workflow file step ids. Confirm a `flow: planning` issue against a workflow file with `step: plan` advances correctly, and the reverse (`flow: plan` against `step: planning`) also walks. Look for cases where the alias is applied on one seam but not the other.

2. **Evaluate-step skip semantics with non-canonical intervening steps.** When the workflow has `evaluate → lint-pass → plan → …` and complexity is `simple`, the walker scans forward for the first non-`plan` step → finds `lint-pass` (non-canonical) → returns null. Is this the right behavior, or should the walker continue past non-canonical ids? Same concern with `complex` — advancing by one to a non-canonical step returns null. Does this surprise users who add intermediate steps?

3. **Legacy fallback boundary.** When `workflow.source.isLegacy === true` but `steps:` happens to exist (synthesized kickoff step from the legacy fallback ladder), the fallback path runs the hardcoded matrix instead of the walker. Does this hide a real configuration error from the user? Should `flowAdvanceDecision` emit a diagnostic when it falls back, so the user knows their per-project WORKFLOW.md isn't on the modules schema yet?

4. **Race between `loadWorkflowFile` and `metadataCache`.** `advanceFlowAndLaunch` calls `loadWorkflowFile` (which reads frontmatter via `metadataCache`) right after `setFlow` mutates the issue's `flow:`. The workflow file itself isn't being mutated, but verify there's no scenario where `loadWorkflowFile` reads a stale `metadataCache` entry for the workflow file (e.g. just after `op-edit-workflow` writes a new version mid-session).

5. **Scaffold-seeded workflow + composition pipeline.** The seeded WORKFLOW.md has all-empty `modules: []` step lists. Per OP-197 / OP-199, an empty modules list composes to no injection. Verify `composeWorkflow` and the kickoff-fallback path in `composeWorkflowSection` don't error on an empty modules list — particularly when the issue is at the `evaluate` step which has `modules: []`.

6. **Loosened `setFlow` validation security/UX.** Now that `flow:` accepts any non-empty trimmed string, can a URI value land something problematic in frontmatter (e.g. `flow:\n  type: evil`)? `processFrontMatter` writes a string scalar so YAML injection is bounded by Obsidian's frontmatter writer, but verify `op-set-flow` URI handling doesn't accept embedded newlines that could break frontmatter on round-trip.

7. **`nextFlow` field rename impact.** The `FlowAdvanceOutput` type adds `nextStep` and keeps `nextFlow` as a wire-format alias. Verify no consumer in `main.ts` (or any test) reads `nextFlow` expecting the legacy enum names (e.g. `planning`); the field always carries canonical ids now even when the input flow was legacy.

8. **Cross-project step semantics.** A user can declare `evaluate → custom-step-A → review` (skipping plan + implement). Does the walker handle that gracefully? `flow: evaluate + complexity: complex` would advance by one to `custom-step-A` (non-canonical) → null. Is the behavior surprising for workflows that intentionally elide canonical steps?

Issue #216, opened 2026-04-26
develop migration from WORKFLOW.md to composeable workflow modules driven by plugin


earchibald:
Tracked as op issue **OP-181**.

## Scope

- [ ] This issue will develop, then drive, migration from WORKFLOW.md to a central, editable plugin-managed set of composeable workflow MODULES. Workflow modules will be used to dynamically inject chunks of workflow according to a composable schema to deployed agent sessions in a given order, allowing us to reuse modules across projects. Preferred coding agent and model should be able to be provided for the entire workflow (the workflow default) and for different steps of the composed workflow, allowing us to implement a dynamic skillset for workflow intent.
- [ ] Additionally, because we have the plugin engine to handle things, modules will TEMPLATE against variables (plugin-provided always-on variables, or variables defined by the user in plugin (global and per-project) .
- [ ] Research: should composed modules be presented to agents as skills or as raw prompt matter? If the latter, do we break what is injected down dynamically across the portions of the workflow or all up front? Recomend these ideas for NOW, LATER or REJECT. If NOW, include in the plan for implementation.
- [ ] Our prior work on agent flows should be integrated into and ported to this idea, making composable workflow modules our standard and retiring the old agent flows implementation and UX.
- [ ] Please research both our codebase and how you want to implement this migration plan under the hood very carefully, using subagents to manage context. The product of this issue will be the scoped plan, broken down into an ordered series of child issues. As appropriate, those child issues may themselves be the parents of child issue/s.
- [ ] In THIS issue, you will document the safe, verifiable implementation path for this very large change and lay out the issues for efficient implementation, allowing parallel child issues to work as needed but keeping our safety intact. Everything should be tested using OP-Test vault (but of course tracked in Agent-Vault!) so that we don't break core functionality while we work.
- [ ] Based on this description, add or reflect back to me suggestions for adding things that help flesh out this idea that I have not covered.

PR #277, opened 2026-04-26
OP-187: op-export-module / op-import-module / op-undo-last-import

earchibald:
## Summary

Child 4 of the [OP-181 workflow-modules umbrella](https://github.com/earchibald/obsidian-projects/issues/216). Three thin convenience wrappers around the plain-markdown module file format, plus the import-time bootstrap of missing vars and one-step undo described in `docs/plans/OP-181-workflow-modules.md` §"Export / import — modules are just markdown".

- **`op-export-module id=<id> | project=<slug>`** — writes `Projects/_op-export/<id>.md` (single) or `Projects/_op-export/<slug>/<id>.md` (bundle); preserves every `VarDecl` form (bare / default-shorthand / object).
- **`op-import-module path=<vault-or-abs-path>`** — parses + plans + commits. For each `{{vars.X}}` body reference, walks the precedence chain (Project → Global → Module-default); missing vars trigger either a per-var modal prompt (palette) or an `ok:false status:"needs-input"` payload with each missing var's `prefill`/`hasModuleDefault` (URI/CLI). Vars land in `settings.workflowVars` (global) or `Projects/<slug>/STATUS.md vars:` (project). Existing target files are backed up to `Projects/_op-import-history/<ts>.bak/`. Transaction record at `Projects/_op-import-history/<ts>.json`.
- **`op-undo-last-import`** — reverses the latest transaction (one-step only): trashes imports, restores backups, prunes only the vars this import added (`preexisting=true` rows are preserved), trashes the tx + bak so the next undo is a no-op. `status:"no-history"` (not an error) when nothing to undo.

Architecture follows the existing `*Pure.ts` / `*.ts` split: `exportImportPure.ts` owns all schema-shape decisions; `exportModule.ts` / `importModule.ts` / `undoLastImport.ts` own vault reads/writes. URI senders accept both `var.<name>=<value>` per-key form and packed `vars=name=val\nname2=val2` (matches OP-204's `parseLaunchVarsFromUri` pattern).

**47 new tests, 1494 / 1494 full suite green.** op-obsidian bumped 0.74.0 → 0.75.0 (minor — additive, no breaking change).

## Test plan

- [x] `npx vitest run` — full suite green (1494/1494)
- [x] `npx tsc --noEmit` — no new errors (8 pre-existing on main, unchanged)
- [x] `npm run build` — main.js rebuilt
- [x] `node scripts/dev-sync.mjs` — copied to OP-Test, plugin reloaded; all three commands registered
- [x] Smoke round-trip in OP-Test:
  - export `id=orient` → wrote `Projects/_op-export/orient.md`
  - delete source → import (no var) → `needs-input` payload with `prefill: "git status"` from inline default
  - re-dispatch with `var.preflight='git status -uno'` → module landed, `settings.workflowVars.preflight === "git status -uno"`, transaction record written
  - `op-undo-last-import` → orient gone, `workflowVars: {}`; second invocation returned `status: "no-history"`
  - overwrite branch: import v1 → import v2 (overwrote=true, backupPath set, var marked preexisting=true); undo of v2 restored backup + preserved v1's var
  - project-bundle export wrote `Projects/_op-export/<slug>/<id>.md`

## Adversarial review pressure-tests for @copilot

- **needs-input contract**: does the URI/CLI surface every missing var with the right `prefill` (especially the empty-string case for bare vars vs a real default)?
- **Backup path collision**: what if two imports happen in the same second? Filename is `YYYYMMDD-HHmmss` — does anything misbehave?
- **Project-slug rewrite**: when `scope=project` but the bundle's `project:` already matches the target slug, does `originalProject === rewrittenProject` cause any wasted work?
- **Empty-string answer**: the spec says `name=` is an explicit empty default. Does `varAnswers: { foo: "" }` round-trip correctly through commit + undo?
- **STATUS.md vars: map**: when the project's STATUS.md has no `vars:` field, does the import correctly add it (without disturbing other frontmatter)? When undo prunes the only var, does it remove the now-empty map?
- **Concurrent overwrite**: import v1 → import v2 → import v3 → undo. Does undo correctly reverse only v3 (not v3+v2)?
- **Corrupted history file**: a stray non-JSON file in `Projects/_op-import-history/`. Does undo skip past it to the next-most-recent valid record?
- **Frontmatter quoting**: a module with `title: "Edge: case"` or a `vars: [foo=2026-04-26]` (Date-coercible). Does the export-import round trip survive YAML auto-coercion?

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.74.1",
+  "version": "0.75.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.75.1",
+  "version": "0.76.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.75.1",
+  "version": "0.76.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.74.1",
+  "version": "0.75.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/cliHandlers.test.ts
+++ b/plugins/op-obsidian/src/cliHandlers.test.ts
@@ -15,6 +15,8 @@ import {
   parseGetSkillParams,
   parseExplainWorkflowParams,
   parseListVarsParams,
+  parseExportModuleParams,
+  parseImportModuleParams,
 } from "./cliHandlers";
 
 describe("parseWorkParams", () => {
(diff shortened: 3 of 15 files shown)


earchibald:
@copilot Please pressure-test this PR adversarially. Focus areas:

1. **needs-input contract roundtrip**. URI/CLI returns `ok:false status:"needs-input"` with `needsInput.vars[]`. Verify each entry's `prefill` matches the planner's `pickPrefill` (default-shorthand → `value`; object form with `default` → that string; bare → `""`). Empty-string `prefill` for bare vars must NOT be mistaken for "no prompt needed".

2. **Empty-string answers**. `var.foo=` (empty value) is an explicit answer per the spec ("name=" is distinct from bare). Does the planner accept it as a real answer (no prompt re-issued)? Does the transaction record write `value: ""`? Does undo prune it correctly?

3. **STATUS.md vars: map mutation**. `applyVarWrite` calls `processFrontMatter` and writes `fm.vars = {...existing, [name]: value}`. (a) If STATUS.md has no `vars:` field, is one created without disturbing other frontmatter? (b) On undo, when removing the only var leaves an empty map, does the code `delete fm.vars` (yes — see undoLastImport.ts) — does that correctly omit the field rather than write `vars: {}`?

4. **Backup path collision**. Two imports in the same wall-clock second produce identical `YYYYMMDD-HHmmss` filenames. The `app.vault.create` of the second tx record will throw because the file already exists. Should we add a uniqueness suffix, or accept the throw + Notice as the "correct" failure mode for human-cadence usage?

5. **Project-slug rewrite identity**. If bundle's `project: foo` and `targetProjectSlug = "foo"`, `rewrittenProject === originalProject`. Does the tx record bother writing both? Does anything downstream assume "rewritten implies different"?

6. **Concurrent overwrite chain**. Import v1 → v2 → v3, then undo. Undo only the latest tx (v3). Does the v3 backup contain v2 (correct: v3 backed up what was on disk before v3 landed = v2)? After undo, is v1's transaction still at the most-recent slot for the next undo? Does the cascading-undo actually work?

7. **Corrupted history file**. A non-JSON `.json` (or a `.json` with unknown `version: 99`) in `Projects/_op-import-history/`. `findLatestTransaction` should skip and try the next-most-recent — does it? Does it ever surface an error if EVERY record is corrupted?

8. **YAML round-trip safety**. The export quoter quotes anything starting with `-?\d` plus a curated unsafe-char regex. Test cases: `title: "2026-04-26"` (Date-coercible), `default: "true"` (bool-coercible), `default: "@bob"` (starts with `@`), `default: ""` (empty), `default: "  leading-trailing  "` (whitespace). Does each survive `format → parseImportFile → re-format` byte-identically?

9. **Source path: vault-relative vs absolute**. If `path=` is vault-relative AND begins with `/` (vault path containing a leading slash quirk), what happens? The current `readImportSource` treats `startsWith("/")` as "skip vault lookup" — is that always correct?

10. **Tests use `vi.mock("obsidian", ...)`**. The `Modal` class isn't exported by my mock for `importModule.test.ts` because I inlined `readProjectVars` to break the cascade. Verify this didn't break anything subtle: did I lose any behavior by inlining vs importing the canonical `readProjectVars` from `explainWorkflow.ts`?

Find specific bugs, propose fixes, push commits if you spot something concrete. Don't just LGTM — turn the screws.

PR #278, opened 2026-04-26
OP-209: update agent-facing docs for workflow-modules cutover

earchibald:
## Summary

- Bundled op-get-skill manual gains a "Workflow modules — what gets injected at launch" section covering file layout (`Projects/_op-modules/`, `Projects/<slug>/MODULES/`, `Projects/<slug>/WORKFLOW.md`, the global `extends:` target), the two template-var namespaces (plugin vars from the always-on registry, user vars `{{vars.<name>}}` declared by modules and resolved through a four-layer precedence chain), how `scope:` is metadata while routing happens via `steps[].modules`, per-step agent/model selection (with a worked YAML example covering scalar/list/keyed-map shapes), the OP-181 visibility tenet for `launchHeadlessSubtask`, and the verify/share verbs.
- Plugin commands table grows to document `op-edit-module`, `op-explain-workflow`, `op-list-vars`, `op-export-module`, `op-import-module`, `op-undo-last-import`. `op-edit-workflow` re-described as the editor entry point for the modern workflow file format. Each row matches the actual handler contract (verified by reading the params parsers + payload builders in `cliHandlers.ts` / `*Pure.ts`).
- `op-work` step 2 reframed to call out that under `workflowMode: "modules"` the kickoff prompt is already composed at launch and `op-explain-workflow` is the diagnostic surface to verify what was injected. Notes that the shipped default is currently `"legacy"`; OP-208 flips it separately, so the prose is merge-order-agnostic.
- `plugins/op/skills/op/SKILL.md` (stub): line refreshed to mention the workflow-modules verbs alongside the existing op-* surface.
- `CLAUDE.md`: added an `op-explain-workflow` + `op-list-vars` smoke-probe step in §4 alongside the existing dev:console + `executeCommandById` checks; cross-references `node scripts/smoke-workflow-modules.mjs` for non-trivial changes.
- `README.md`: workflow modules now headlined under "What's inside" with a one-sentence pointer to `docs/workflow-modules/` and `docs/specs/workflow-{module,file}-schema.md`; "Why the split?" §5 reworded to describe the workflow-modules system as the way projects express SDLC opinions. Distinguishes plugin vars from user vars so the four-layer precedence claim only applies to user vars.
- Plugin version bumped 0.76.0 → **0.76.3** across the three lockstep files (manifest.json + package.json + plugin.json). Three patches in this PR: initial doc landing (0.76.1), first review-fix pass (0.76.2), second review-fix pass (0.76.3).

Lands paired with **OP-208** (8a — flip `workflowMode` default to `modules` + retire legacy "Workflow inline cap" row); together they close OP-191's umbrella scope. Distinct from OP-193 (Child 10) human-facing docs at `docs/workflow-modules/` — this PR is the in-skill, agent-instruction-grade mirror.

## Test plan

- [x] `npm test -- --run skill` — 8 / 8 pass (validates the skill loader still finds and loads the bundled manual).
- [x] `npm test -- --run docsExamples` — 6 / 6 pass.
- [x] `node scripts/bump-version.mjs <ver>` rebuilt `main.js` fresher than `manifest.json` (OP-105 freshness guardrail) for each iteration.
- [x] `node scripts/dev-sync.mjs` synced into OP-Test; plugin reload reported each new version.
- [x] `obsidian vault=OP-Test op-get-skill` returns the updated manual with all new keywords present.
- [x] **Two rounds of adversarial Copilot review** caught 10 places where prose drifted from shipped behavior (var namespaces, scope-as-routing, op-explain/list-vars/export/import/undo contracts, workflowMode default). All fixed against the implementation.


earchibald:
Pushed aaabca1 fixing all five review findings:

1. **op-explain-workflow row** — dropped fictional `model` / `launchVars` optional params; corrected payload description to match `buildExplainPayload` shape (`composed.text`, `composed.chunks`, `vars` with per-var `scope`, `diagnostics`); dropped the "active workflow path" claim.
2. **op-list-vars row** — rewrote to describe its actual job (browsing `PLUGIN_VAR_REGISTRY` with optional context-resolved `currentValue`), dropped unsupported `agent` param, pointed at `op-explain-workflow` for user-var precedence introspection.
3. **Four-layer precedence section** — split into two namespaces: plugin vars (`{{id}}`, `{{today}}`, …) come from the registry and are not layered; only user vars (`{{vars.<name>}}`, declared in module `vars:` frontmatter) flow through the four-layer chain. Fixed the example syntax — `vars: [name=value]` inline / `vars:\\n  - name=value` block (no `- ` inside the inline brackets).
4. **scope: section** — retitled "`scope:` is metadata, not step routing". Step routing is via the workflow file's `steps[].modules` list; `scope:` is metadata used for chunk records + the intra-scope-collision check on user-var declarations.
5. **README "What's inside" bullet** — distinguished plugin vars from user vars so the four-layer precedence claim only applies to user vars.

Plugin bumped 0.76.1 → 0.76.2; manual rebuilt + dev-synced + skill tests still 8/8 green.

earchibald:
Pushed a0ee352 fixing the second-pass review:

1. **op-export-module row** — corrected to match the real contract: `id` **or** `project` (mutually exclusive), no `module`/`dest`/`scope`/`include_used_workflow`. Writes to fixed `Projects/_op-export/`.
2. **op-undo-last-import row** — replaced fictional `noop` return with the real `status: "ok"` / `status: "no-history"` shape and the CLI's `no transaction history to undo` message.
3. **op-import-module row** — tightened var-answer wording to `vars=name=val\\n…` packed or `var.<name>=value` per-key (the actual CLI surface).
4. **"workflowMode default" wording (two places)** — dropped the "(default since OP-208)" claim. Reworded both spots so the manual is correct regardless of which PR merges first: states "shipped default is currently legacy; OP-208 flips it."
5. **PR description version mismatch** — updated to reflect 0.76.3 (three patches in this PR: 0.76.1 initial, 0.76.2 first fix pass, 0.76.3 this fix pass).

Plugin bumped 0.76.2 → 0.76.3, manual rebuilt + dev-synced, skill tests green.

earchibald:
Pushed 9d73144 (third review fix) + 896dd78 (merge of origin/main with OP-208 landed):

**9d73144** — third review pass fix:
- "Sharing modules across vaults" rewritten to match actual export/import behavior: `id=<id>` writes single `.md`; `project=<slug>` writes per-project subfolder; modules only (no workflow-file bundling); imports overwrite with backup under `Projects/_op-import-history/<ts>.bak/` instead of being "surfaced as diagnostics".

**896dd78** — merge in OP-208 cutover (just landed) + version reconciliation:
- OP-208 (#279) flipped `DEFAULT_SETTINGS.workflowMode` from `legacy` to `modules` while OP-209 was in review. Merged in. Conflicts on the three lockstep version JSONs resolved by landing on **0.77.1**.
- Updated the merge-order-agnostic prose in the manual back to a positive statement: `modules` is the post-OP-208 default for fresh installs (the conditional "shipped default until OP-208 flips it" wording is no longer needed).

Tests: skill 8/8, build green, dev-sync into OP-Test reported 0.77.1.

Should be the last round — every Copilot finding to date has been addressed against the implementation.

PR #279, opened 2026-04-26
OP-208: cutover — flip workflowMode default to modules + retire legacy injection cap row

earchibald:
## Summary

- **Default flip:** `DEFAULT_SETTINGS.workflowMode` is now `'modules'`. `mergeSettings` preserves explicitly-saved `'legacy'` values; absent-field blobs pick up the new default.
- **Legacy promptBuild blob removed:** the workflow injection branch in `buildPrompt` no longer reads `WORKFLOW.md` verbatim under a `From <path>:` preamble. Both modes route through `composeWorkflowSection`, whose own legacy-fallback ladder (`workflowFile.ts` shapes 1/2/3/5) wraps vanilla `WORKFLOW.md` into a synthetic kickoff step. The `workflowMode` arg-driven gate inside `buildPrompt` is gone (the `main.ts` evaluator gate too).
- **Settings cap row retired:** the `Workflow inline cap (chars)` row in the Injection collapsible is deleted (replaced by the Workflows top-level group from OP-201). The `maxWorkflowChars` field stays on `OpSettings` because the composer / editor validator still honor it as a soft size budget.

## Migration semantics

| Saved data.json | Loaded `workflowMode` | Behavior |
| --- | --- | --- |
| Field absent | `"modules"` (new default) | Modules engine; vanilla `WORKFLOW.md` handled by legacy-fallback ladder |
| `"legacy"` (explicit) | `"legacy"` (preserved) | Modules engine still runs in `promptBuild` (legacy code path removed); `launchAgentModal` still shows "modules disabled" notices — those gates are out of scope here (belong to OP-209's docs/agent-facing pass) |
| `"modules"` | `"modules"` (preserved) | Unchanged |

Stored-`'legacy'` users functionally use the modules engine after this PR. The legacy-fallback ladder preserves their user-visible kickoff content for vanilla `WORKFLOW.md`.

## Test plan

- [x] `npm test` — 1530/1530 vitest cases pass (one pre-existing 0-test suite failure in `iTermPrefs.test.ts` from a Vitest-4 / `obsidian` package resolution quirk, unrelated to this PR — confirmed by running the same test against `main`).
- [x] Settings test additions: `DEFAULT_SETTINGS.workflowMode === "modules"`; `mergeSettings({ workflowMode: "legacy" })` preserves `"legacy"`; absent-field blob → `"modules"`.
- [x] promptBuild test rewrite: vanilla `WORKFLOW.md` routes through the legacy-fallback ladder; explicit `workflowMode: "legacy"` is now byte-identical to the implicit default (gate is dead in `buildPrompt`).
- [x] Smoke test against `OP-Test` vault: plugin loads at v0.77.0; live `workflowMode === "modules"`; "Workflow inline cap (chars)" row absent from the Injection collapsible; "Include project workflow" toggle still present; `dev:console` clean.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.76.0",
+  "version": "0.77.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.76.0",
+  "version": "0.77.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/main.ts
+++ b/plugins/op-obsidian/src/main.ts
@@ -1899,53 +1899,52 @@ export default class OpPlugin extends Plugin {
         statusLine: (line) => notify(`op-evaluate ${entry.id}: ${line}`),
         paneStream: (chunk) => console.log(`[op-evaluate ${entry.id}] ${chunk}`),
       });
-      // OP-199 (2b): in modules-mode, compose the workflow's `evaluate` step
-      // and prepend it to the evaluator's prompt. Launch context is best-
-      // effort here (no `wd` resolution because the evaluator runs in-
-      // process, not in a terminal) — `repoPath` from `resolveRepoPath`,
-      // `branch` via `git rev-parse` at that path. Evaluator-step modules
-      // that reference `{{branch}}` etc. resolve to either the real value
-      // or a `missing-var` diagnostic, never break the launch.
-      const composeWorkflowSectionFn =
-        this.settings.workflowMode === "modules"
-          ? async () => {
-              const profile = resolveProfile(this.settings, this.settings.defaultAgent);
-              const repoPath = resolveRepoPath(this.app, this.settings, entry.project);
-              const branch = repoPath ? await gitBranchAt(repoPath) : undefined;
-              const parentRaw = (this.app.metadataCache.getFileCache(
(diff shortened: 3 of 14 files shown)


earchibald:
@copilot please pressure-test this OP-208 cutover PR with the following:

1. **Migration semantics correctness.** `mergeSettings` (`settingsPure.ts` ~L344) is supposed to preserve `workflowMode: 'legacy'` for users who explicitly opted in, while flipping absent-field blobs to the new `'modules'` default. Trace the validator: does it correctly handle `workflowMode: undefined` (explicit), `workflowMode: ""`, `workflowMode: null`, and the absent-field case? Are there edge cases (e.g. arrays/objects coerced via `WORKFLOW_MODES.has`) that could silently corrupt the migration?
2. **Legacy fallback removal in `promptBuild.ts`.** I dropped the `else { read WORKFLOW.md verbatim }` branch on the assumption that `composeWorkflowSection`'s legacy-fallback ladder (`workflowFile.ts` shapes 1/2/3/5 → `synthesizeLegacyWorkflow`) covers vanilla WORKFLOW.md. Is there any vanilla-WORKFLOW.md shape that the composer's classifier would reject (e.g. shape 4 — frontmatter parses to wrong type) where the user previously got a useful "From <path>:" inline blob and now gets nothing? Should the dropped path-only fallback (when `body.length > maxWorkflowChars`) be preserved somewhere, or is the composer's `size-budget` info diagnostic sufficient?
3. **Evaluator gate removal.** `main.ts` `runEvaluatorForIssue` previously only composed when `workflowMode === 'modules'`. Now it always composes. For a stored-`'legacy'` user, the evaluator will now spend the `EVALUATOR_MAX_WORKFLOW_CHARS`-clamped budget on workflow content they didn't ask for. Acceptable tradeoff or do we need a corresponding settings UI nudge?
4. **`BuildPromptArgs.workflowMode` is now dead in `promptBuild`** but still passed by callers (`openAgent.ts`, `main.ts` evaluator). Worth removing entirely or keeping it for type safety with `launchAgentModal`'s UI gates?
5. **Version bump classification.** Bumped to `0.77.0` (minor — additive default change). Is `minor` appropriate, or should the default flip be classified `major` since it changes prompt-injection behaviour for fresh installs and stored-`'legacy'` users alike?
6. **Test coverage gaps.** I dropped the byte-identical legacy reference test. Is there a missing case in the rewritten `promptBuild.test.ts` that would catch a regression in the legacy-fallback ladder's WORKFLOW.md handling at the `buildPrompt` integration layer?

PR #282, opened 2026-04-27
OP-179: iTerm tab title — append [Parent: <PARENT-ID>] suffix when issue has a parent

earchibald:
## Summary
- Surface `parent:` on `IssueEntry` (parser + type), plumb `parentId` through `openAgent` → `LaunchArgs` → `OrchestrateArgs` → `BuildViewArgs`.
- New `labelWithParent(label, parentId)` helper in `orchestrator.ts` appends ` [Parent: <PARENT-ID>]` to the OSC 1 / OSC 2 / `setSessionName` payloads when set; `oscSafe` is applied to the parent id alongside the id/title.
- 10 new vitest cases (`labelWithParent` + `buildViewScript` parent coverage). Full suite green: 1541 / 1541.
- UI affordance reuses the existing canonical writer `op-set-link issue=<id> relation=parent target=<PARENT>`. Palette command and settings field are follow-ups.
- Builds on OP-177's id-prefix work; no behavior change for non-child issues.

## Test plan
- [x] `npx vitest run` in `plugins/op-obsidian/`
- [x] `node scripts/bump-version.mjs` → 0.78.0 (minor — additive `parent:` surfacing)
- [x] `node scripts/dev-sync.mjs` to OP-Test, plugin reload OK
- [x] OP-Test smoke: `IssueEntry.parent` flips through `processFrontMatter` round-trip; deployed `main.js` contains the `[Parent: ${o}]` literal and the `parentId` path through `orchestrateLaunch`; console clear

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.78.0",
+  "version": "0.78.1",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.77.1",
+  "version": "0.78.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.77.1",
+  "version": "0.78.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.78.0",
+  "version": "0.78.1",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/issueStore.ts
+++ b/plugins/op-obsidian/src/issueStore.ts
@@ -175,6 +175,7 @@ export class IssueStore extends Component {
       githubIssue: str(fm.github_issue),
       agent: str(fm.agent),
       title: str(fm.title) ?? file.basename,
+      parent: str(fm.parent),
       resolvedFolder: inResolved,
     };
   }
--- a/plugins/op-obsidian/src/openAgent.ts
+++ b/plugins/op-obsidian/src/openAgent.ts
@@ -270,6 +270,12 @@ export async function openAgent(
     tmuxBinary: settings.tmuxBinary,
     issueId: args.entry.id,
     issueTitle: args.entry.title,
+    // OP-179: append ` [Parent: <PARENT-ID>]` to the iTerm tab/window/session
+    // label when this issue has a parent. Sourced from the parsed entry; if
+    // the entry's `parent` is missing (legacy issue not yet re-indexed), fall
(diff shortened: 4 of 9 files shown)


earchibald:
@copilot — please pressure-test OP-179 for behavioral correctness. Concrete prompts to verify against the diff:

**`plugins/op-obsidian/src/issueStore.ts:178` — `parent` parser**
- Claim: `parent: str(fm.parent)` returns `undefined` for non-string YAML values (e.g. multi-parent arrays `[OP-1, OP-2]`), mirroring `readParentId`'s posture in `openAgent.ts`. Verify by tracing `str()` in this file: does it actually reject arrays/numbers, or could a YAML array silently coerce to its first element?

**`plugins/op-obsidian/src/openAgent.ts` — parentId source-of-truth precedence**
- Claim at the `launchInTerminal({…, parentId: …})` site: the parsed `args.entry.parent` is the primary source, with `readParentId(app, args.entry.path)` as the cache-lag fallback. Verify the chain `args.entry.parent ?? readParentId(app, args.entry.path) ?? undefined`:
  - When `entry.parent === ""` (empty string, not undefined), what wins? `??` only short-circuits on null/undefined, so empty-string `entry.parent` would be passed through untouched and `labelWithParent` would treat it as "no suffix" — does that match the intent, or should the `openAgent` site coalesce empty strings to undefined too?
  - When `readParentId` returns `null`, does `null ?? undefined` correctly yield `undefined`? (Yes per spec, but worth confirming the exact path through.)

**`plugins/op-obsidian/src/orchestrator.ts:88-91` — re-attach path label composition**
- Claim: the re-attach path (where the iTerm session still exists from a prior launch) recomputes `sessionTitle` with the current `parentId` and re-applies it via `setSessionName`. Verify: if a user sets `parent:` on an in-flight issue *after* the agent's pane was created, does a subsequent `op:open-agent` invocation actually update the live tab title, or does iTerm cache the OSC payload such that a re-attach `setSessionName` is a no-op? (The orchestrator stores the prior `sessionTitle` only via `iTerm/driver.setSessionName`; tmux's view script OSC payload only fires on cold attach.) State the observable behavior.

**`plugins/op-obsidian/src/orchestrator.ts:333-337` — `buildViewScript` label composition**
- Claim: `oscSafe` is applied to `parentId` *before* the `[Parent: …]` suffix is composed, so a control char injected into `fm.parent` (e.g. `OP-149\x07injected`) cannot terminate the OSC payload. Verify by tracing the code path. Also verify `labelWithParent` itself does not re-run `oscSafe` (avoiding double-strip is fine; the concern is whether anywhere in the chain a raw `parentId` reaches a `printf '\033]…;%s\007'` payload unstripped).

**`plugins/op-obsidian/src/orchestrator.ts` — `labelWithParent` semantics**
- Claim: `labelWithParent(label, parentId)` returns `label` unchanged when `parentId` is `undefined` *or* empty string; appends ` [Parent: <PARENT-ID>]` otherwise. Verify edge cases:
  - whitespace-only parent (e.g. `"   "`): does it append `" [Parent:    ]"` (probably yes — current code only checks `!parentId || parentId.length === 0`)? Should it `.trim()` first to match `readParentId`'s `raw.trim().length > 0` guard, for consistency?
  - parent equal to issueId (self-link, e.g. `parent: OP-179` on OP-179): does this slip through and produce a useless `OP-179 title [Parent: OP-179]`? Does the link-graph plumbing reject self-links upstream so this can't happen in practice (cite the relevant `op-set-link` guard)?

**OP-177 dedup interaction**
- Claim: `labelWithId`'s prefix-dedup (when title already starts with `<id>`) is unchanged — `labelWithParent` composes on top. Verify a basename-fallback issue ("OP-179 iTerm tab title…" with no `fm.title`) still produces `OP-179 iTerm tab title… [Parent: OP-149]` rather than `OP-179 OP-179 iTerm tab title… [Parent: OP-149]`.

**OP-178 hook interaction**
- Claim: the OP-178 `window-unlinked` hook is keyed on `tmuxWindow` (i.e. `tmuxWindowName(issueId)`), which is *not* affected by the new suffix — the hook continues to fire only when the agent's specific tmux window dies. Verify the format predicate `#{==:#{hook_window_name},${tmuxWindow}}` does not pick up the new suffix anywhere.

**Tests**
- `orchestrator.test.ts` has 10 new cases. Are there obvious behavioral edges not covered? Specific candidates: parent equal to issueId, whitespace-only parent, parent containing tmux-special chars (`:`, `'`, `\`), parent rendered alongside the OP-178 hook (verify both coexist in the emitted script).

**No-behavior-change for non-child issues**
- Claim: when `parent` is undefined throughout the chain, the OSC payloads, `setSessionName` payloads, and tmux window name are byte-identical to OP-177's output. Verify by spot-checking the existing test fixtures still pass without modification.

Push fixes for any inaccuracies you find. I'll re-pull, re-verify, and merge once findings are addressed.

PR #291, opened 2026-04-27
OP-220: shared slugify util + {{slug}} plugin var + real template tokens in quickstart

earchibald:
## Summary

- Extract `plugins/op-obsidian/src/slug.ts` exporting `slugify(input, opts)` with caseFold / allowUnderscore / maxLen / stripLeadingTaskPrefix / fallback knobs. Migrate `tmuxWindowName` onto it as a thin preset; existing `terminalLaunch.test.ts` cases pass byte-identical (regression contract).
- Add `{{slug}}` to `PLUGIN_VAR_REGISTRY` (Issue identity group). Derived from `title` with `{ caseFold: true, maxLen: 40, stripLeadingTaskPrefix: true }`; chains `|| undefined` so empty titles preserve the existing `missing-var` soft-fail.
- Switch the workflow-modules quickstart `branching.md` example from `<issue-id>-<slug>` prose to real `{{id}}-{{slug}}` tokens; update §4 launch-preview block accordingly. Add `{{slug}}` row to `05-variables-and-templating.md` Issue identity table.

## Open design questions (deferred)

- **Co-resolution with `{{branch}}` once the worktree exists.** Proposed default: re-derive from title (deterministic, pre-worktree-safe). Alternative: parse the branch and infer the slug from it. Not decided here — `{{slug}}` is title-derived; `{{branch}}` retains its existing pre-worktree-undefined semantics.
- **Numeric/lettered task-prefix stripping (`NN[a-z]?:`).** Implemented as opt-in via `stripLeadingTaskPrefix: true` and only triggered for the `{{slug}}` registry entry. Regex `/^\s*\d+[a-z]?\s*:\s*/i` matches `10:`, `10b:`, `1A:` with optional whitespace. Doesn't strip issue-id prefixes (`OP-220 Title`) since they have no colon.
- **Worktree branch construction.** Currently produces `worktree-{{id}}` (no descriptive tail). Whether to include `{{slug}}` there — making branches like `worktree-OP-220-add-slug-plugin-var-extract-shared` — is deferred to a follow-up. Flagged here but not implemented.

## Test plan

- [x] `npx vitest run` — full op-obsidian suite green (1595/1595).
- [x] New `slug.test.ts` covers empty / all-punctuation / unicode / length-cap (with last-`-`-boundary truncation) / leading task-prefix on+off / `allowUnderscore` on+off / `caseFold` on+off / fallback / the registry preset.
- [x] Existing `terminalLaunch.test.ts:15-67` passes unchanged through the migration (regression contract).
- [x] `pluginVarRegistry.test.ts` extended with `{{slug}}` cases — fixture title resolution, prefix stripping, soft-fail contract, length-cap behavior.
- [x] `listVarsPure.test.ts` golden snapshot updated for the new entry order.
- [x] Smoke: `obsidian vault=OP-Test op-list-vars project=demo issue=DM-1` (against op-obsidian@0.80.0) — title "probe chip" → `slug = "probe-chip"`.

### DIFF
--- a/docs/workflow-modules/02-quickstart.md
+++ b/docs/workflow-modules/02-quickstart.md
@@ -101,14 +101,21 @@ order: 10
 ---
 
 Always create a feature branch off `main` before making changes. Branch
-names use `<issue-id>-<slug>`. Push commits to that branch only —
-never directly to `main`.
+names use `{{id}}-{{slug}}`. Push commits to that branch only — never
+directly to `main`.
 ```
 
 That's it: three lines of body, six lines of frontmatter. The filename
 basename (`branching.md`) must match the `id:` field
 (`id: branching`) — a mismatch is a `malformed-frontmatter` diagnostic
 and the module is silently dropped.
+
+`{{id}}` and `{{slug}}` are always-on plugin vars: `{{id}}` is the
+issue's canonical id (e.g. `OP-194`); `{{slug}}` is a kebab-cased,
+40-char-capped slug derived from the issue's `title:`. Together they
+produce a branch suggestion the agent can use verbatim. The full
+registry — and the precedence rules for user-declared vars — lives in
+[05-variables-and-templating.md](./05-variables-and-templating.md).
 
 ## 4. See it in the launch preview (60 seconds)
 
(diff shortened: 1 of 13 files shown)


earchibald:
@copilot adversarial review please. Concrete pressure-test prompts:

1. **`tmuxWindowName` regression contract.** The migration replaces an inlined `[^A-Za-z0-9_-]+ → -, collapse, trim, fallback "agent"` regex chain with `slugify(input, { allowUnderscore: true, fallback: "agent" })`. Stress-test the new path with: empty string, single dash, multiple consecutive separators, mixed-case + digits + underscore + dash, embedded `:`/`.`/`+`, leading & trailing dashes, very long input. Are there any inputs where the old chain and the new util diverge by even one character?

2. **`maxLen` boundary truncation.** The util truncates at the last `-` boundary inside the cap when one exists, otherwise truncates flat. Edge cases: cap = 0, cap = 1, dash at index 0 (sliced to empty → re-trimmed → potentially empty → fallback fires?), cap exactly at a `-` boundary, cap one short of a `-` boundary, all dashes input. Does any input slip past the `lastDash > 0` guard and produce a leading/trailing `-` survivor?

3. **`stripLeadingTaskPrefix` regex precision.** `/^\s*\d+[a-z]?\s*:\s*/i` matches `10:`, `10b:`, `1A:`. Does it over-match? Try `1234567890b: long`, `99999: x`, `1: 2: 3` (multi-colon — does it strip just the first segment?), `: hello` (no digits — should NOT strip), `1.5: x` (dot in prefix — should NOT strip per the regex). Does it under-match a legitimate task-prefix form used in this repo?

4. **Soft-fail contract on the registry.** The registry entry returns `slugify(...) || undefined`. Trace: title is `""`, `"   "`, `"!!!"`, undefined slot in `Partial<RenderContext>`. Each should result in the renderer leaving `{{slug}}` verbatim and firing `missing-var` — not silently producing `""` or `null`.

5. **Quickstart doc accuracy.** §3 now claims `{{id}}` and `{{slug}}` "are always-on plugin vars" with `{{slug}}` "capped at 40 chars on a `-` boundary". §4 claims rendered output is "~190 chars" for an example issue titled "Add module variables". Does the prose match what the implementation actually does for a representative kickoff launch? Anything misleading to a first-time reader?

6. **Listed character-class invariant.** Are there ASCII inputs where the four `(caseFold × allowUnderscore)` combinations produce surprising outputs? E.g., `caseFold=true, allowUnderscore=true` should preserve underscore but lowercase letters; my reading is the regex `/[^a-z0-9_-]+/g` after a `toLowerCase()` is correct, but I want a fresh pair of eyes on whether unicode lowercase (e.g., Turkish dotted-i, German ß) could surface a regression.

PR #350, opened 2026-05-03
OP-225: shared worktreeBranchName helper + slug tail in branch convention

earchibald:
## Summary

- New `plugins/op-obsidian/src/worktreeBranch.ts` exporting `worktreeBranchName(issueId, title)` that consumes the existing `{{slug}}` preset (`caseFold: true, maxLen: 40, stripLeadingTaskPrefix: true`) — single source of truth for the `worktree-<id>[-<slug>]` convention.
- Falls back to bare `worktree-<id>` (no trailing dash) when the title slugs to empty (all-punctuation, whitespace-only, undefined).
- Lifts the visible examples to the new form: `pluginVarRegistry.ts` `branch.example`, `__fixtures__/integration/composed.snapshot.md`, `integration.test.ts` fixture, `docs/workflow-modules/05-variables-and-templating.md` example column. Auto-gen `reference/plugin-vars.md` regenerated via `check-workflow-docs.mjs` (OP-220 CI gotcha).

## Scope notes

The OP-225 spec listed `terminalLaunch.ts` / `orchestrator.ts` as call sites per the OP-220 audit. A codebase scan shows no plugin code today actually constructs `worktree-{{id}}` — branch creation happens externally (EnterWorktree, manual `git worktree add`). The new helper exists so future construction sites have one source of truth; the visible-example surfaces consume the new convention now. `tmuxWindowName` keys on `{{id}}` only via its own `slugify(id, …)` preset and is unchanged (regression contract preserved).

`agentHooks.ts:158`'s advisory stderr message uses `-b $OP_ISSUE_ID` with no `worktree-` prefix at all (pre-existing inconsistency); the bash hook lacks the title at execution time and the issue scopes this out under "no other call sites today".

## Test plan

- [x] `worktreeBranch.test.ts` — normal title → `worktree-<id>-<slug>`; all-punctuation / whitespace / empty / undefined → bare `worktree-<id>`; leading `NN[a-z]?:` stripped; 40-char tail cap on `-` boundary.
- [x] `pluginVarRegistry.test.ts` + `integration.test.ts` updated fixtures pass.
- [x] Full plugin suite: 1741 tests pass.
- [x] `check-workflow-docs.mjs`: re-run, regenerated `reference/plugin-vars.md` committed.
- [x] Smoke: dev-sync into OP-Test, `op-list-vars project=testing issue=TST-5` confirms the new `branch.example` resolves correctly. (Workflow-modules seed not initialized in this env — same state as OP-220 notes; live probe sufficient.)

### DIFF
--- a/docs/workflow-modules/05-variables-and-templating.md
+++ b/docs/workflow-modules/05-variables-and-templating.md
@@ -48,7 +48,7 @@ generated reference docs) reads from there.
 | **Repo / vault** | `{{repo_path}}` | Absolute path to the project's code repo; undefined for meta-only projects. | `/Users/you/Projects/obsidian-projects` |
 |  | `{{vault_path}}` | Absolute path to the active vault. | `/Users/you/work/Agent-Vault` |
 |  | `{{vault_name}}` | Display name of the active vault. | `Agent-Vault` |
-|  | `{{branch}}` | Git branch the agent is on; undefined before the worktree exists. | `worktree-OP-212` |
+|  | `{{branch}}` | Git branch the agent is on; undefined before the worktree exists. | `worktree-OP-220-add-slug-plugin-var-and-extract-shared` |
 | **Run context** | `{{today}}` | ISO `YYYY-MM-DD`, computed by the launching surface. | `2026-04-26` |
 |  | `{{agent}}` | Agent id launching this session. | `claude` |
 |  | `{{model}}` | Resolved model id; undefined when the agent doesn't pick one per launch. | `claude-opus-4-7` |
--- a/docs/workflow-modules/reference/plugin-vars.md
+++ b/docs/workflow-modules/reference/plugin-vars.md
@@ -54,7 +54,7 @@ The issue id is {{id}} — branch is {{branch}}, today is {{today}}.
 | `{{repo_path}}` | `/Users/you/Projects/obsidian-projects` | Absolute path to the project's code repository — undefined for meta-only projects. |
 | `{{vault_path}}` | `/Users/you/work/Agent-Vault` | Absolute path to the active Obsidian vault. |
 | `{{vault_name}}` | `Agent-Vault` | Display name of the active Obsidian vault. |
-| `{{branch}}` | `worktree-OP-194` | The git branch the agent is operating on — undefined before the worktree exists or for meta-only projects. |
+| `{{branch}}` | `worktree-OP-220-add-slug-plugin-var-and-extract-shared` | The git branch the agent is operating on — undefined before the worktree exists or for meta-only projects. |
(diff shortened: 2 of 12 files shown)


earchibald:

1. `plugins/op-obsidian/src/worktreeBranch.ts:13-19` — `worktreeBranchName()` interpolates raw `issueId` into the git ref without any sanitization or validation. An ID containing `/`, whitespace, or other ref-hostile characters would produce a nested or invalid branch name, and this exported helper is positioned as the future source of truth for branch construction. **[for worker]**
2. `plugins/op-obsidian/src/agentHooks.ts:156-159` — the PreToolUse refusal text still tells users to create `-b $OP_ISSUE_ID`, which diverges from the new `worktree-<id>-<slug>` convention introduced in this PR. I left it alone because it lives in the hook/edit-guard path the review brief said not to patch. **[for worker]**
3. `plugins/op-obsidian/src/pluginVarRegistry.ts:173-178`, `plugins/op-obsidian/src/integration.test.ts:164-177` — the newly surfaced branch example and integration fixture were still hardcoded, so a future tweak to the shared slug preset could desync them from `worktreeBranchName()` silently. I fixed that by deriving the integration fixture from the helper and adding a parity assertion for the registry example. **[fixed in 80040c7]**
4. `plugins/op-obsidian/src/worktreeBranch.test.ts:33-42` — the length-cap coverage only asserted `startsWith()` / `<= 40`, which would let some truncation regressions slip through. I tightened it to pin the exact dash-boundary result and added the no-dash flat-truncation edge case. **[fixed in 80040c7]**

earchibald:
## Worker reply

- **Finding #1 (issueId sanitization in `worktreeBranchName`)** — **waived.** Issue IDs in the schema are `[A-Z]+-\d+` (enforced by `op-new` filename sanitization and frontmatter shape) and are already branch-safe. Adding sanitization inside this helper would mask schema bugs upstream rather than catch them. The helper trusts the schema invariant the same way `slugify` and `tmuxWindowName` do.

- **Finding #2 (`agentHooks.ts:158` advisory)** — **fixed in `978b375`.** Aligned to `-b worktree-$OP_ISSUE_ID`. Slug isn't available at hook-execution time, so this is a prefix-only alignment; the worker still has full freedom to extend the branch name with a slug tail when running in their own worktree.

- **Findings #3 (parity assertion + helper-derived fixtures) and #4 (tightened length-cap test)** — accepted as `80040c7`. Pulled into the worktree, reran the affected suites: 63/63 pass.

Both copilot fix commits + the worker advisory fix are smoke-tested at `op-obsidian@0.86.2` (dev-sync into OP-Test, plugin reloaded, `op-list-vars` still resolves the new `branch.example` cleanly).

PR #357, opened 2026-05-03
OP-241: port-input inline error + lifecycle smoke

earchibald:
## Summary

- Inline error feedback on the Dashboard Port input: out-of-range or non-numeric values now badge the input (`aria-invalid`, `.op-port-input--invalid`) and render a `.op-port-error` span; valid input clears it. Empty input (mid-edit) is treated distinctly so the badge doesn't flash while the user is clearing the field.
- New `scripts/smoke-settings-lifecycle.mjs` — CLI-driven DOM probe that opens the Settings tab, expands the Dashboard section, exercises the port-input badge round-trip, and verifies close+reopen re-renders are stable. Mirrors the existing `smoke-workflow-modules.mjs` harness shape per CLAUDE.md §"Smoke-test Settings UIs".

Closes the OP-235 review nits captured as OP-241 scope.

## Test plan

- [x] `npx vitest run` — full suite (1747/1747) including new `validateDashboardPortInput` cases.
- [x] `node scripts/bump-version.mjs minor` (0.86.3 → 0.87.0) — build green, three version files in lockstep.
- [x] `node scripts/dev-sync.mjs` — synced into OP-Test.
- [x] `node scripts/smoke-settings-lifecycle.mjs` — PASS (60 rows × 4 lifecycle renders, badge round-trip OK).
- [x] `obsidian vault=OP-Test dev:console` — no errors after the smoke run.


earchibald:

1. `plugins/op-obsidian/src/settings.ts:1872-1875` — The new inline error node mixed `role="alert"` with `aria-live="polite"`, which gives conflicting live-region semantics and inconsistent announcement timing across screen readers. I dropped the explicit `aria-live` and left the self-contained alert role. [fixed in 0d4ccff]
2. `plugins/op-obsidian/src/settingsPure.test.ts:98-141` — The validator behavior was correct, but the new test coverage missed the adversarial inputs called out here: whitespace-padded values, leading zeros, `-0`, `+1024`, scientific notation, Arabic-Indic digits, and oversized digit strings. I added regression coverage for those cases. [fixed in a0972a7]
3. `scripts/smoke-settings-lifecycle.mjs:79-204` — The first cut of the smoke harness only compared total `.setting-item` row counts across 4 renders, so a duplicate `.op-port-error` / `.op-port-input` node could slip through if the overall row count stayed stable. I hardened it to dispatch both `input` and `change`, assert exactly one port input + one error node, and stretch the lifecycle loop to 10 renders. [fixed in 6052baa]
4. `plugins/op-obsidian/src/settings.ts:237-245`, `plugins/op-obsidian/src/settings.ts:331-359`, `plugins/op-obsidian/src/settings.ts:1859-1863` — **Waived:** I did not find a DOM-node leak across re-renders. `display()` empties the tab container, `rerenderSection()` empties the section wrapper, and `renderDashboard()` aborts the previous controller before wiring the new subtree, so the new `portError` div is recreated instead of duplicated. [for worker]
5. `plugins/op-obsidian/src/settingsPure.ts:488-500` — **Waived:** persisted-but-invalid `dashboard.port` values are already coerced away during `mergeSettings`, so the Port input does not boot into an unbadged invalid state from corrupted `data.json`. [for worker]
6. `plugins/op-obsidian/src/settings.ts:1854-1863`, `plugins/op-obsidian/src/settings.ts:2000-2012` — **Waived:** the OP-235 `AbortController` invariant still holds after the port-error addition. The only async dashboard probe still uses the fresh controller’s signal and bails before DOM writes when aborted; the new port-validation path is synchronous and does not retain the prior controller. [for worker]
7. `scripts/smoke-settings-lifecycle.mjs:66-76`, `scripts/smoke-settings-lifecycle.mjs:174-183`, `CLAUDE.md:44-57` — **Waived:** the harness does rely on internal Obsidian Settings APIs (`app.setting.openTabById`, `app.setting.close`), but that is already the repo-sanctioned recipe for Settings-tab smoke tests and is acceptable for a local, non-CI harness. [for worker]
8. `plugins/op-obsidian/src/settingsPure.ts:95-115` — **Waived:** `/^\d+$/` intentionally restricts the validator to ASCII decimal digits. That rejects Unicode digits, `+1024`, `1e3`, and `0x1234`; those are reasonable rejections here, and they now have explicit regression coverage. Whitespace-padded valid values still pass because the input is trimmed first, and leading zeros normalize to the same in-range port. [for worker]
9. `plugins/op-obsidian/manifest.json:1-7`, `plugins/op-obsidian/package.json:1-10`, `plugins/op/.claude-plugin/plugin.json:1-7`, `docs/specs/OP-217-agent-dashboard.md:318-320` — **Waived:** the PR’s `0.86.3 -> 0.87.0` minor bump is defensible. This is additive user-visible UX plus a new smoke harness, and the dashboard spec’s own release recipe treats additive surface as a minor bump. My follow-up review fixes bumped that to `0.87.1`. [for worker]

PR #360, opened 2026-05-03
OP-242: op-dashboard daemon — implement /regenerate-token + /restart, wire data.json config

earchibald:
## Summary

- **Daemon** (`op-dashboard.py`): implements `POST /regenerate-token` (rotate token, persist via `write_token`, broadcast WS close 4401 to live clients) and `POST /restart` (set `state.restart_requested` + `state.shutdown_event`; `run()` execs `sys.executable` with the same argv after `runner.cleanup()`). Both gated on the *current* `X-Op-Token` so a malicious local process can't rotate the token without already holding it.
- **Plugin install**: `installDaemon()` now takes an optional config and writes `op-dashboard.config.json` next to the daemon. Both Setup-modal and Settings-tab install paths pass `{vaultDataPaths: [<vault>/.obsidian/plugins/op-obsidian/data.json]}` so the daemon's `model`/`workdir`/`started_at` enrichment is populated on first install instead of staying empty.
- **Plugin UI**: drops the staged-work `404` fallback strings on Restart and Regenerate now that both endpoints exist.
- **Spec**: OP-217 §AC #9 updated to describe the implemented endpoint and its 4401 broadcast.
- **Version**: `0.87.2 → 0.88.0` (minor — additive endpoints + new install behaviour).

## Test plan

- [x] `python3 dashboard/test_op_dashboard.py` (47 → 52 tests, all OK with iTerm-bundled python that has aiohttp; aiohttp-gated tests cover auth, token rotation invalidating a live WS with close 4401, restart setting the flag + shutdown event, and `run()` exec'ing with `sys.executable` + `sys.argv` after cleanup).
- [x] `npx vitest run` — 1760 tests pass (3 new tests in `dashboardSetupModal.test.ts` cover the config sidecar shape, the omit-when-no-config path, and `pluginDataJsonPath`).
- [x] OP-Test smoke: `dev-sync` + plugin reload reports v0.88.0; settings tab opens cleanly; `app.vault.adapter.basePath` resolves so Settings → Install will write the config sidecar.
- [ ] **Live daemon path** (next install/upgrade): confirm `~/Library/Application Support/iTerm2/Scripts/AutoLaunch/op-dashboard.config.json` lands with the expected shape after clicking Install in Settings, and confirm Settings → Regenerate flips the open dashboard tab to "invalid token". The daemon-side endpoint is covered by unit tests; the end-to-end click is the remaining manual step.


earchibald:

Local `copilot --autopilot --allow-all` ran against the diff. It pushed one fix commit (`a2cd7b9`) and surfaced these findings; my dispositions are inline.

### Pushed fixes (kept)

1. **`Hub.close_all` snapshot pruning** (`op-dashboard.py:548–571`). After awaiting `ws.close()` on each client, prune just the snapshot from `self._clients`. Per-handler `finally: hub.remove(ws)` already runs so double-discard is harmless, but pruning here avoids stale entries when `ws.close()` raises or the handler hasn't unwound yet. New `HubTests.test_close_all_prunes_snapshot_clients_even_when_close_raises` covers the failure path with a fake WS that raises on `.close()`. [fixed in `a2cd7b9`]
2. **Malformed-config defensive test** (`test_op_dashboard.py`, `DataJsonReaderTests.test_find_data_json_paths_ignores_malformed_config`). Daemon already swallows the parse exception (`find_data_json_paths` wraps the read in try/except), this just locks the behavior in. [fixed in `a2cd7b9`]

### Pushed fix re-reverted

3. **404 fallback notices in `settings.ts`** — copilot reintroduced "Restart endpoint not yet implemented…" / "Regenerate endpoint not yet implemented…" to handle the mixed plugin/daemon-version case. Issue scope explicitly required removing them; the new plugin's Install button writes the matching daemon, so a stale daemon recovers in one click. Re-reverted in `86aed85` with rationale. [reverted in `86aed85`]

### Open `[for worker]` items waived

4. **`pluginDataJsonPath` hardcodes `.obsidian`.** Most users are on the default config dir; `app.vault.configDir` would be more correct for users with a custom config dir (`OBSIDIAN_CONFIG_DIR` / portable installs). Codebase already hardcodes the same path elsewhere (`main.ts:1610`, `dashboardInstall.ts:29`). Out of scope for OP-242; tracking as a follow-up if it ever bites a real user.
5. **`/restart` concurrent POSTs.** Two simultaneous POSTs both see `state.restart_requested=True` and both return 200. That's harmless — the shutdown event fires once via `call_soon`, `runner.cleanup()` runs once, `os.execv` runs once. The second response is just truthful: a restart was requested, you got 200.
6. **`?token=` vs `X-Op-Token` for mutating POSTs.** Both transports route through the same `request_token` + `token_ok` check. The header form keeps tokens out of access logs (the daemon's logger doesn't log query strings, but referrer leaks are still a pattern worth noting); the plugin's UI uses the header form on its `fetch()` calls already. No change needed.
7. **`os.execv` from `iterm2.run_forever` autolaunch path.** `execv` replaces the entire process image — fds, sockets, the iterm2 connection are all torn down by the kernel when the new image takes over. iTerm's AutoLaunch supervisor sees the new process come up and reconnects on the next tick. Verified covered by `RunRestartExecTests.test_run_execs_when_restart_requested`.
8. **Test reassigns `opd.TOKEN_PATH` inside try/finally.** A test failure before the finally would pollute subsequent tests. The reassignments only happen inside `_run()` coroutines wrapped in `try/finally` — the finally runs even if an assertion raises. Acceptable for now; if these tests grow, switch to `unittest.mock.patch.object(opd, "TOKEN_PATH", …)`.

### Verification

- `python3 plugins/op-obsidian/dashboard/test_op_dashboard.py` — 49 passing (originally 47; +2 from the fix commit).
- `npx vitest run` — 1760 passing.
- OP-Test smoke: `dev-sync` reports plugin loaded at v0.88.2 (`0.87.2 → 0.88.0` for the feature, `0.88.0 → 0.88.1` for the copilot fix bump, `0.88.1 → 0.88.2` for the 404-revert).

earchibald:

1. [fixed in a2cd7b9] `plugins/op-obsidian/dashboard/op-dashboard.py:555-572` — `Hub.close_all()` closed a snapshot of live sockets but previously relied on each handler's later `finally: hub.remove(ws)` to clear the registry. If `ws.close()` raised, or the handler unwound slowly, the hub could retain dead clients and keep broadcasting to them. I now prune only the closed snapshot and added a regression test in `plugins/op-obsidian/dashboard/test_op_dashboard.py:294-318`.
2. [fixed in a2cd7b9] `plugins/op-obsidian/src/settings.ts:1969-1975,2072-2077` — the Settings UI dropped its 404-specific fallback copy. After a plugin upgrade talking to an older daemon, Restart/Regenerate regressed from actionable guidance to generic `HTTP 404`. I restored the explicit degraded-path messages.
3. [fixed in a2cd7b9] `plugins/op-obsidian/dashboard/op-dashboard.py:429-438` — malformed `op-dashboard.config.json` sidecars were handled defensively in code, but there was no regression coverage for that path. I added one in `plugins/op-obsidian/dashboard/test_op_dashboard.py:231-248` so a future parser refactor can't silently break the fallback.
4. [for worker] `plugins/op-obsidian/dashboard/op-dashboard.py:668-680,702-710` — the file/token ordering is fine for new handshakes (the token file is rewritten and `state.token` flips before the first `await`), so an old-token reconnect does **not** get back in. The remaining gap is already-open sockets: between `state.token = new_token` and each awaited `ws.close()`, an old client can still dispatch frames because admission is only checked once in `ws_handler()` and message dispatch never revalidates the token.
5. [for worker] `plugins/op-obsidian/dashboard/op-dashboard.py:689-693,1027-1029` — `/restart` still depends on `loop.call_soon(state.shutdown_event.set)` being "late enough" for the `200` to flush before `runner.cleanup()` tears the server down. That may be acceptable, but making the flush explicit changes the shutdown choreography and needs a deliberate decision/test under a real aiohttp runtime.
6. [for worker] `plugins/op-obsidian/src/dashboardSetupModal.ts:485-497` — `pluginDataJsonPath()` hardcodes `.obsidian/plugins/op-obsidian/data.json` off `adapter.basePath`. Vaults using a non-default Obsidian config dir will write a stale sidecar path; this should key off `app.vault.configDir` (or an equivalent API), which touches vault-path handling.
7. [for worker] `plugins/op-obsidian/dashboard/op-dashboard.py:577-581,668-685` — mutating POSTs accept both `?token=` and `X-Op-Token`. The UI correctly uses the header, but keeping URL-token auth enabled for state-changing endpoints is a security/API choice because query tokens are easier to leak into logs/history than custom headers.
8. [for worker] `plugins/op-obsidian/dashboard/op-dashboard.py:683-693` — concurrent `/restart` calls are effectively idempotent and every caller still gets `200` once `restart_requested` is already true. That's probably okay, but changing it would be a public-surface choice (`200` vs `202`/`409`), so I left it alone.

Waived after source review:

- `plugins/op-obsidian/dashboard/op-dashboard.py:1029-1035` — the `execv`/AutoLaunch path did not show an obvious code defect: aiohttp cleanup happens before exec, and the remaining FD-leak question depends on runtime/library behavior rather than a clear bug in this diff.
- `plugins/op-obsidian/dashboard/test_op_dashboard.py:467-570` — the aiohttp-gated tests already restore `TOKEN_PATH` in `finally` blocks, so ordinary assertion failures do not pollute later tests.

PR #374, opened 2026-05-03
OP-260: Phase 3 new-file pretool guard layer (default-off)

### LINKED ISSUE #365: Phase 3: new-file pretool guard layer (default-off)
Tracked as op issue **OP-260**.

## Scope

Phase 3 of OP-218 (parent: OP-257; depends on Phase 2 ergonomics being shaken out, but technically can start independently). Refuse agent `Write` of new files under `Projects/<slug>/{ISSUES,RESOLVED ISSUES,TASKS}/` — creation must go through `op-new` / `op-task-create`.

- [ ] Extend `agentHooks.ts` `writeGuardScript` with a new-file refusal layer (after Phase 2's managed-note layer). Trigger when `tool_input.file_path` resolves to a vault path matching the protected folder pattern AND the file does not yet exist.
- [ ] Add `OP_ALLOW_NEW_FILE=1` escape-hatch env var.
- [ ] Refusal message points at `op-new` for issues, `op-task-create` for tasks (no message for RESOLVED ISSUES — that folder is plugin-managed at resolve time and direct creation there is always wrong).
- [ ] New plugin setting `agentDiscipline.newFileGuard: bool`, **default off**.
- [ ] `agentHooks.pretool.test.ts` cases for: (a) refused new ISSUES path, (b) refused new TASKS path, (c) escape hatch, (d) existing file modify still flows through Phase 2's layer, (e) other paths under `Projects/` still allowed.
- [ ] Manual smoke against OP-Test.
- [ ] Phase 6 flips the default — do NOT flip in this PR.

earchibald:
## Summary

- Adds a third PreToolUse layer for op-launched agents that refuses agent `Write` of new files under `Projects/<slug>/{ISSUES,RESOLVED ISSUES,TASKS}/`. Refusal hint points at `op-new` for issues and `op-task-create` for tasks; `RESOLVED ISSUES` refuses without a command hint (plugin-managed at resolve time). Override per-call via `OP_ALLOW_NEW_FILE=1`.
- Default off behind a new `agentDiscipline.newFileGuard` setting (third toggle in the worktree-enforcement section). Phase 6 of OP-218 owns the flip — **do not flip in this PR**.
- Refactors Layers 2 and 3 to share a single stdin parse prefix. Layer 2's earlier `exit 0` short-circuits drained stdin before Layer 3 could read it; both layers now use control flow instead of early exits when emitted together. Existing-file edits still flow through Layer 2 first; new-file refusal only fires when the path matches the protected folder pattern AND the file does not yet exist.

## Test plan

- [x] `agentHooks.pretool.test.ts` covers the spec's (a)–(e) cases plus opt-in/opt-out script-body assertions and an all-three-layers-off uninstall path. 29/29 pass.
- [x] OP-Test smoke: toggle on → Layer 3 emitted, installed script refuses a new ISSUES path with the right hint, accepts under `OP_ALLOW_NEW_FILE=1`, accepts paths outside `Projects/`. Toggle off → Layer 3 dropped from the script body.
- [x] Settings tests untouched (20/20 pass on `settingsPure`).

Parent: OP-257 (Phase 2: OP-259). Closes #365.


earchibald:

All tests pass (29/29 in `agentHooks.pretool.test.ts`). One mechanical fix pushed; remaining findings flagged for the worker.

---

### 1. `schema.md` missing Phase 3 documentation [fixed in 2360ebb]

`plugins/op/skills/op/reference/schema.md:276` had a thorough Phase 2 (OP-259) guard-layer entry but no matching entry for Phase 3. `OP_ALLOW_NEW_FILE`, `agentDiscipline.newFileGuard`, the shared stdin-parse design, and independent per-layer overrides were all absent. Added a Phase 3 paragraph immediately after the Phase 2 entry.

---

### 2. Layer 2 refactor — observable behavior unchanged [waived]

The old Layer 2 code short-circuited `exit 0` when `OP_ALLOW_MANAGED_EDIT=1` before consuming stdin. The new code moves stdin parsing to a shared block above both layers and uses a negated outer condition instead.

Walk of every scenario (L2 only unless noted):

| OP_ALLOW_MANAGED_EDIT | stdin | Old exit | New exit | Same? |
|---|---|---|---|---|
| `1` | any | 0 | 0 (L2 block skipped) | yes |
| unset | empty | 0 | 0 | yes |
| unset | managed `.md` | 2 | 2 | yes |
| unset | unmanaged `.md` | 0 | 0 | yes |
| L1+L2, main checkout | any | 2 (L1) | 2 (L1) | yes |

One new interaction old code could never produce: **L2+L3 both enabled, `OP_ALLOW_MANAGED_EDIT=1`, new file under ISSUES/** → L2 skipped, L3 fires, exit 2. This is intentional and correct — the two overrides are independent. Verified by direct bash test.

---

### 3. Unanchored glob `*/Projects/*/ISSUES/*` [for worker]

`agentHooks.ts:288` — the Layer 3 `case` pattern matches any path segment chain containing `/Projects/<x>/ISSUES/`. This repo has a `docs/examples/` tree with `Projects/` subdirs (currently no ISSUES/TASKS children, but the structure exists). An op-launched agent writing `docs/examples/.../Projects/tutorial-project/ISSUES/DEMO-3 new.md` would hit Layer 3 even though it's a fixture file.

Risk is low today (guard gates on `$OP_ISSUE_ID`; `docs/examples` has no ISSUES dirs). Worth documenting the assumption or adding an exclusion before Phase 6 flips the defaults on.

---

### 4. `case` backslash-space for RESOLVED ISSUES [waived]

`agentHooks.ts:288` — `*/Projects/*/RESOLVED\ ISSUES/*`. Verified with a live bash `case` execution against a literal-space path — matches correctly. bash `case` treats `\ ` as a literal space in glob patterns.

---

### 5. `[ ! -e "$file" ]` with missing parent directory [waived]

Verified: `[ ! -e "/nonexistent/parent/file.md" ]` returns true regardless of whether the parent exists. POSIX `test -e` on a path with a missing component returns false, not an error. Guard correctly fires when parent dir is absent.

---

### 6. Hint string double-quoting survival [waived]

`agentHooks.ts:291` emits `new_hint="Use op-new project=<slug> title=\"...\" to create new issues..."`. Inspected the installed script body directly — `\"` inside a double-quoted assignment is valid bash and produces `title="..."` at echo time. No escaping bug.

---

### 7. Test (d) — cross-layer isolation proof [for worker, minor]

`agentHooks.pretool.test.ts:448` — Test (d) uses an existing managed file, so `[ ! -e "$file" ]` is false and Layer 3 would be a no-op regardless. The assertion `.not.toMatch(/refusing creation/)` proves a counterfactual. Comment at line 457 explains this clearly, so intent is documented.

**Missing coverage:** no test for `OP_ALLOW_MANAGED_EDIT=1` + new file under ISSUES + both layers on → Layer 3 should still fire. Suggested addition:

```ts
it("OP_ALLOW_MANAGED_EDIT=1 does NOT bypass Layer 3 on new files", async () => {
  const note = path.join(wt, "Projects", "demo", "ISSUES", "DEMO-5 new.md");
  await fs.mkdir(path.dirname(note), { recursive: true });
  const r = await runGuard(
    { OP_ISSUE_ID: "DEMO-5", OP_ALLOW_MANAGED_EDIT: "1" },
    wt, payload(note));
  expect(r.code).toBe(2);
  expect(r.stderr).toMatch(/refusing creation of new file/);
});
```

---

### 8. `OP_ALLOW_NEW_FILE` documentation [waived]

Mentioned in `agentHooks.ts` header, `settingsPure.ts` JSDoc, `settings.ts` UI description, and schema.md (fixed in finding 1). README does not document any `OP_ALLOW_*` vars — consistent with `OP_ALLOW_MAIN_EDIT` and `OP_ALLOW_MANAGED_EDIT`. No drift relative to existing guards.

---

### 9. `HookInstallResult` / `wantInstall` uninstall path [waived]

`agentHooks.ts:91` — `wantInstall = enforceWorktree || managedNoteGuard || newFileGuard`. When all three are false the guard is uninstalled. Test "uninstalls the guard when all three layers are off" (line 519) covers this.

---

### 10. Settings UI placement [waived]

New toggle added inside `renderWorktreeEnforcement` after the OP-259 toggle. Section heading "Worktree enforcement" already houses Layers 1 and 2; Layer 3 is consistent. No heading drift.

earchibald:
## Worker reply to adversarial review

Thanks. Findings handled:

- **#1** — accepted (already pushed in 2360ebb).
- **#2, #4, #5, #6, #8, #9, #10** — waived findings reviewed and concur.
- **#3** (unanchored `*/Projects/*/ISSUES/*` glob) — acknowledged. The `docs/examples/.../Projects/*/` fixtures currently have no `ISSUES`/`TASKS` children, so the false-positive surface is empty today. Tightening this to vault-anchored requires plumbing the vault path through env into the bash script — out of scope for Phase 3 (default-off). Will revisit in **Phase 6** before the default flips on, where it has to be addressed alongside the soak feedback.
- **#7** — fixed in 68ec85b. Added an `OP_ALLOW_MANAGED_EDIT=1 + new file` isolation test confirming Layer 3 fires independently of Layer 2's env override (30/30 pass).

PR is green and ready to merge.

PR #390, opened 2026-05-05
OP-265: pass vault=<name> in op skill CLI guidance + record vault in STATUS.md

### LINKED ISSUE #372: Always pass vault=<name> in op skill's automatic CLI guidance to launched agents
Tracked as op issue **OP-265**.

## Scope

- [ ] - The op skill (and obsidian-ops-specialist subagent) currently does not instruct launched agents to pass an explicit vault selector when invoking the obsidian CLI.
- [ ] - Result: when the user has multiple vaults registered and Obsidian foregrounds a different vault
- [ ] agent CLI calls silently target the wrong vault. Real failure observed: 'op-append-commit issue=PRC-2' returned 'Issue not found' because OP-Test was active in Obsidian even though Agent-Vault was the intended target. State corruption is possible (e.g. mutations land in wrong vault).
- [ ] - Fix: bake 'always pass vault=<active-project-vault-name> on every obsidian CLI call' into:
- [ ] - The top-level op skill markdown (the 'Plugin commands' section + 'Delegating vault/CLI work' note).
- [ ] - The obsidian-ops-specialist subagent's system prompt / description.
- [ ] - Any example commands in the skill (op-scaffold
- [ ] op-new
- [ ] op-work
- [ ] op-append-commit
- [ ] op-set-pr
- [ ] op-set-scope
- [ ] op-resolve
- [ ] eval
- [ ] property:set
- [ ] read).
- [ ] - Bonus: op-scaffold should record the resolved vault name into Projects/<slug>/STATUS.md so subsequent /op:* invocations can derive the vault selector deterministically without re-probing.
- [ ] - Acceptance: a fresh agent run that opens a different vault in Obsidian still mutates the intended project vault correctly because every CLI call carries vault=<name>.

earchibald:
## Summary

- Bake `vault=<name>` into every example obsidian invocation across the op skill surface (manual.md, SKILL stub, obsidian-ops-specialist subagent, cli-gotchas property recipe). Each surface now names the failure mode the rule prevents (`op-append-commit issue=PRC-2` returning "Issue not found" because the wrong vault was foregrounded) so the guidance is unmissable.
- `op-scaffold` (and the firstRunReadme demo path) now writes `vault: <app.vault.getName()>` into the new `STATUS.md` frontmatter — launched agents can derive the CLI selector deterministically from STATUS.md without re-probing `obsidian vault`. Pre-OP-265 projects keep working via the discovery-call fallback.
- `plugins/op-obsidian` 0.94.3 → 0.95.0 (additive user-facing field).

Closes https://github.com/earchibald/obsidian-projects/issues/372.

## Test plan

- [x] `npx vitest run` — 1857 / 1857 pass.
- [x] `obsidian vault=OP-Test op-get-skill` after dev-sync — content carries `obsidian vault=<name>` for op-scaffold / op-new / op-work / op-set-section / op-append-commit / op-set-pr / op-set-link / op-link-check / op-resolve / op-explain-workflow.
- [x] `obsidian vault=OP-Test op-scaffold slug=vault-probe prefix=VPP` writes `vault: OP-Test` into STATUS.md (verified, then torn down).
- [x] `obsidian vault=OP-Test dev:console` clean post-smoke.

### DIFF
--- a/plugins/op-obsidian/manifest.json
+++ b/plugins/op-obsidian/manifest.json
@@ -1,7 +1,7 @@
 {
   "id": "op-obsidian",
   "name": "Obsidian Projects (op)",
-  "version": "0.94.4",
+  "version": "0.95.0",
   "minAppVersion": "1.5.0",
   "description": "Jira-lite issue tracker for Obsidian vaults — deterministic commands backing the op workflow.",
   "author": "Eugene Archibald",
--- a/plugins/op-obsidian/package.json
+++ b/plugins/op-obsidian/package.json
@@ -1,6 +1,6 @@
 {
   "name": "op-obsidian",
-  "version": "0.94.4",
+  "version": "0.95.0",
   "description": "Obsidian plugin that owns the deterministic half of the op workflow.",
   "main": "main.js",
   "scripts": {
--- a/plugins/op-obsidian/src/firstRunReadme.ts
+++ b/plugins/op-obsidian/src/firstRunReadme.ts
@@ -105,13 +105,16 @@ export function buildDemoStatusBody(): string {
   ].join("\n");
 }
 
-/** Frontmatter for the demo STATUS.md note. */
-export function demoStatusFrontmatter(): string {
+/** Frontmatter for the demo STATUS.md note. OP-265: `vault:` records the
+ *  active vault name at scaffold time so launched agents can derive the
+ *  `vault=<name>` CLI selector from STATUS.md without re-probing. */
+export function demoStatusFrontmatter(vault: string): string {
   return [
     "---",
     "project: " + DEMO_PROJECT_SLUG,
     "type: project-status",
     "prefix: " + DEMO_PROJECT_PREFIX,
+    "vault: " + vault,
     "tags:",
     "  - project/" + DEMO_PROJECT_SLUG,
     "  - status",
@@ -281,7 +284,7 @@ export async function scaffoldDemoProject(app: App): Promise<{
   const issuesDir = `${DEMO_PROJECT_FOLDER}/ISSUES`;
   await ensureFolder(app, DEMO_PROJECT_FOLDER);
   await ensureFolder(app, issuesDir);
-  await app.vault.create(statusPath, demoStatusFrontmatter() + buildDemoStatusBody());
+  await app.vault.create(statusPath, demoStatusFrontmatter(app.vault.getName()) + buildDemoStatusBody());
   await app.vault.create(`${DEMO_PROJECT_FOLDER}/${DEMO_PROJECT_SLUG}.base`, demoBaseBody());
   for (const seed of DEMO_ISSUES) {
     const filename = `${DEMO_PROJECT_PREFIX}-${seed.number} ${seed.title}.md`;
--- a/plugins/op-obsidian/src/scaffoldProject.ts
+++ b/plugins/op-obsidian/src/scaffoldProject.ts
@@ -73,7 +73,11 @@ export async function scaffoldProject(
   await app.vault.create(basePath, renderBase(slug));
 
   const statusPath = normalizePath(`${projectFolder}/STATUS.md`);
-  await app.vault.create(statusPath, renderStatus(slug, prefix, repoPath));
+  // OP-265: record the resolved vault name so launched agents can derive the
+  // `vault=<name>` selector deterministically from STATUS.md without
+  // re-probing `obsidian vault`. Source of truth at scaffold time is
+  // app.vault.getName() — whichever vault Obsidian is currently writing to.
+  await app.vault.create(statusPath, renderStatus(slug, prefix, app.vault.getName(), repoPath));
 
   // OP-188: seed the default workflow file so the orchestrator's
   // workflow-file walker (flowOrchestrator.ts) finds a canonical step list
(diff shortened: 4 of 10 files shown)


earchibald:

Reviewed against every pressure-test prompt. One fix committed; remaining findings documented below.

---

### 1. SKILL.md: `obsidian op-get-skill` missing `vault=<name>` [fixed in 2576e43]

**File:** `plugins/op/skills/op/SKILL.md` — "Fetching the full manual" code block (was line 30).

The PR correctly updated zero-order rule 1 in the same file to require `vault=<name>` before every invocation, but the very next code block — the `op-get-skill` fetch example — was left vault-less:

```bash
# Before (PR omission)
obsidian op-get-skill

# After (fix)
obsidian vault=<name> op-get-skill
```

An agent following the new rule 1 ("cache the vault, target every call") then hitting the literal fetch example would see a contradiction. Fixed mechanically in 2576e43.

---

### 2. `schema.md` line 150: `property:read` uses `file=` (not `path=`) [for worker]

**File:** `plugins/op/skills/op/reference/schema.md` line 150 (pre-existing, not introduced by this PR):

> fall back to `property:read name=id file=<name>` if needed

All other `property:read` examples throughout this repo use `path=` (cli-gotchas.md line 36, line 52; new manual.md prose line 47). The correct parameter name per cli-gotchas is `path=`. This inconsistency predates PR #390 and isn't in the PR's diff, but schema.md was touched by this PR so it surfaced in scope. Low-risk doc fix; flagging rather than touching unexplored pre-existing prose.

---

### 3. `demoStatusFrontmatter(vault)` signature — no external callers [waived]

**File:** `plugins/op-obsidian/src/firstRunReadme.ts`

Signature change from `demoStatusFrontmatter()` → `demoStatusFrontmatter(vault: string)` breaks only direct call sites. The only call site is line 287 (updated in this PR). The test file (`firstRunReadme.test.ts`) does not import or call `demoStatusFrontmatter` at all — tests cover `scaffoldFirstRunReadme`, `buildReadmeBody`, `buildDemoIssueNote`, etc. No breakage.

---

### 4. `healFrontmatter` does not touch STATUS.md [waived]

`healFrontmatter` classifies files by path regex (`ISSUES_RE`, `RESOLVED_RE`, `TASKS_RE`). STATUS.md matches none of them and is explicitly documented as out of scope (migrateAddManagedFlag.ts line 20). Adding `vault:` to STATUS.md frontmatter causes zero interaction with the heal pass.

---

### 5. `listProjects` / `findProjectByPrefix` / `repoPath` — tolerate new field [waived]

`projects.ts` reads only `fm.prefix` from STATUS.md frontmatter (after checking `fm?.type === "project-status"`). `repoPath.ts` reads only `fm.repo_path`. Neither touches `vault:`. Extra keys in YAML frontmatter are ignored by JavaScript property access. No collision.

---

### 6. Bases / cross-project aggregation [waived]

Demo and scaffold .base files filter by `file.folder.contains("<slug>")` — folder-based, not type-based. STATUS.md is already included in those results today (it lives in the project folder). The new `vault:` field would appear as an extra column in table views if Obsidian Bases auto-surfaces all frontmatter — this is additive and harmless (actually useful for cross-vault views). No aggregation rule breaks.

---

### 7. Bootstrap logic ("Where to source the name") — no chicken-and-egg [waived]

The prose in manual.md correctly describes a two-step bootstrap: (1) `obsidian vault` (no args, intentionally vault-less) to discover registered vaults, (2) cache the name and use `vault=<name>` on all subsequent calls. The STATUS.md `vault:` field is valuable for agents that already have STATUS.md content injected into their kickoff context — they can read `vault:` from the YAML frontmatter without any CLI call. The CLI example (`obsidian vault=<cached> property:read name=vault path=...`) is a valid secondary path (e.g., for agents that have a STATUS.md path but not its contents). No circular dependency.

---

### 8. `obsidian read path=` validity [waived]

`obsidian vault=<name> read path="…"` is a pre-existing pattern in manual.md (appears in Verb: work step 5 and the Notes idempotency check). This PR only added `vault=<name>` to existing `read path=` invocations. The pattern also appears in cli-gotchas.md line 118 as "fall back to `obsidian read`". This is established codebase usage, not a new form introduced by this PR.

---

### 9. `property:read path=` parameter [waived]

The new prose in manual.md line 47 uses `property:read name=vault path="Projects/<slug>/STATUS.md"`. cli-gotchas.md line 36 explicitly documents: "use `obsidian property:read name=<key> path=<vault-relative-path>`" — so `path=` is the correct parameter name. Consistent.

---

### 10. `vault="$VAULT"` with shell quoting [waived]

`obsidian vault="$VAULT" property:read ...` in cli-gotchas.md is valid shell. The shell expands `$VAULT` and passes `vault=<expanded-value>` as a positional argument to obsidian. Double-quoting around the value prevents word-splitting on vault names with spaces. Correct.

---

### 11. manual.md size [waived]

File is 48,067 bytes (~47 KB). No evidence of a size cap in the `op-get-skill` handler or bundled skill infrastructure. The PR adds approximately 1.5 KB of new prose. Not a concern.

---

### 12. Semver 0.94.3 → 0.95.0 (minor) classification [waived]

The only API surface change is `demoStatusFrontmatter(vault: string)` — an exported TypeScript function within the plugin source. It is not part of the plugin's runtime JSON API surface (not a CLI command, not a URI handler). No external consumer can call it. The `vault:` STATUS.md field is purely additive. Minor bump is appropriate; major would be an overclassification.

---

### 13. `docs/workflow-modules/` vault-less examples (informational, out of PR scope)

`docs/workflow-modules/04-faq.md`, `05-variables-and-templating.md`, `04-compose-your-first-workflow.md`, and `migration/from-agent-flows.md` contain vault-less `obsidian op-*` invocations. These files were **not touched by this PR** and predate the vault-targeting rule. They are user-facing documentation (not the op skill manual that agents consume). Flagging for a separate cleanup pass; not a blocker for this PR.

